PR #31, opened 2016-03-12
Ethernetprovider_add: Adding reply_echo_cmd kwarg to ethernet_provider.py

les67:
Added a new kwarg called reply_echo_cmd which is initialized to False. If the ethernet provider sends response back with original command + response terminator, this removes the first bit and returns only the response from the machine. If your machine has this feature, set reply_echo_cmd to True in your config file.


laroque:
This looks pretty good, but you need to update the docstring for __init__ to reflect the new optional argument.

PR #35, opened 2016-04-20
Feature/muxer get spime conf string add

les67:
Added conf_str as an initializing parameter for the MuxerGetSpime class, default is None. Purpose is for endpoints to internally store its configuration string. Tested and works well.

### DIFF
--- a/dragonfly/implementations/prologix.py
+++ b/dragonfly/implementations/prologix.py
@@ -122,7 +122,7 @@ class MuxerGetSpime(SimpleSCPIGetSpime):
     def __init__(self, ch_number, conf_str = None, **kwargs):
 	if conf_str is None:
 		logger.debug('no configuration string given for endpoint')
-	else	
+	else:	
 		self.conf_str = conf_str        
 	self.base_str = "DATA:LAST? (@{})"
         self.ch_number = ch_number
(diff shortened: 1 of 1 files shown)


laroque:
I'm going ahead and merging this. Just a style note, we usually do not put a space around the "=" in an method's signature, but do put them in code execution. This is purely stylistic and has no impact on execution, but it is nice to keep consistent style within the code just so things are easier to read.

Also, you removed that logger.debug statement. I agree that having it as a debug is weird, but may actually want to issue a logger.warning there if no configuration is provided. Is your plan to skip endpoints that don't have a configuration string?

I can see wanting to be able to explicitly skip providing one (maybe you set it manually but don't want to put it in the config file for some reason). This could easily lead to confusion though, because if something was set manually and working, then the machine were factory reset, restarting the service may not actually get things working again.

I'd suggest putting that bit of code back in with a warning statement. Then in your MultiplexerProvider (or whatever you name it), when you loop over the endpoints, I'd say that if the value is `None` (the default you used), then you raise an exception that a configuration value is required, but if it is `False` (as in the boolean, which could be explicitly set in the config file), then the service will leave that channel alone silently (becomes someone has explicitly said to skip it).

les67:
Ditto on the style note. As far as an endpoint missing a conf_str value, yes that was my plan. So if the endpoint was missing a conf_str value it would neither be configured nor added to the scan list. I do agree that it's a good idea to differentiate between an endpoint with an empty value for conf_str and an endpoint with a conf_str value set to False. In this way we can raise an exception for the former and be able to handle to later (i.e. not configure the channel for that endpoint). I will re-add the debug part with a warning for the former case.

PR #64, opened 2016-09-30
Feature/sql snapshot

les67:
Added implementation class SQLSnapshot with methods get_logs and get_latest. The former queries the sc db and returns logs (value_raw, value_cal, and timestamp) for all present endpoints between two specified timestamps (for a given sc db table of this class). Note that the value_cal is a formatted JSON dictionary. The latter returns the latest logged values (value_cal and timestamp) for all endpoints in the passed list.

Generalized GenericAgent's _call to parse pretty-print for all derived classes (Get, Cmd, Set, etc); pretty-print removed from derived classes. Also upgraded cast_arg to use ast.literal_eval. This is helpful in directly parsing in Python types: dict, bool, tuple, list, None in addition to str and numbers.


laroque:
@les67 could you post the config file you've been using for your tests?

les67:
I have only tested against the backup db on Gus. I should also test with the db on Myrna but will have to coordinate with the operator to find a time. The contents of the config I've been using so far are in this text file:
[test_config.txt](https://github.com/project8/dragonfly/files/504505/test_config.txt)

laroque:
I've done a very minimal test of this with [insectarium](../../insectarium) and it seems to work. The steps for that were:
1) clone insectarium
2) enter dragonfly/dragonfly subdirectory and change to this branch for testing
3) docker-compose build with `--no-cache` option on `kv_stores` target to force dragonfly services to rebuild (using the new dragonfly version)
4) attach to the kv_stores container and start an additional process with the config above (changed to point to proper hosts on the container network)
5) attach to the kv_stores container again and run `dragonfly cmd db_num_table.get_latest timestamp=2016-10-03T04:00:00Z endpoint_list="['flux_capacitance']" -b rabbit_broker`

PR #73, opened 2016-11-12
Feature/daq providerwith sql snapshot

les67:
The new SQLSnapshot class is in sql_snapshot.py which defines the methods we need for taking snapshot of the slow control database, i.e. get_latest and get_logs. DAQProvider has been upgraded to use these methods as well; it also includes a new debug flag for rf_roi measurements which is useful when performing virtual runs with insectarium. The class RunDBInterface in daq_db_interface.py has also been upgraded to return fields of type <code>timestamp</code> from the database; this is used when asking for a run's <code>start_timestamp</code> in DAQProvider (for example). All tested in insectarium and working well.


dragonfly/implementations/daq_run_interface.py
```diff
@@ -108,18 +125,34 @@ def start_run(self, run_name):
...
+        logger.info('doing prerun meta-data and snapshot gets')
+        meta_result = self.provider.get(self._metadata_state_target, timeout=120)
+        these_metadata = meta_result['value_raw']
         self._run_meta.update(these_metadata)
-        self.determine_RF_ROI()
+        for target,item_list in self._snapshot_target_items.items():
+            snapshot_result = self.provider.cmd(target, 'get_latest', [self._start_time,item_list], timeout=120)
+            these_snaps = snapshot_result['value_raw']
+            self._prerun_snapshot.update(these_snaps)
+        if not self._debug_without_rf_roi:
```
guiguem: why determine RF_ROI here and not at line 155 too?

dragonfly/implementations/daq_run_interface.py
```diff
@@ -92,8 +104,13 @@ def run_name(self, value):
             return
         result = self.provider.cmd(self.run_table_endpoint, 'do_insert', payload={'run_name':value})
         self.run_id = result['run_id']
+        self._start_time = result['start_timestamp']
 
     def end_run(self):
+        self._postrun_snapshot = {}
+        self._do_postrun_gets()
```
guiguem: do we want to grab the data from the db even if we are in debug mode?

dragonfly/implementations/daq_run_interface.py
laroque: The RF_ROI goes into the metadata file which becomes tag values in DIRAC. That file gets sent to the metadata receiver prior to the start of data taking and may (@nsoblath ?) be responsible for triggering making a directory for the run. Dirac needs this info in order to apply the tags to the data when they are added to the catalog.

Line 155 is data for the post-run snapshot, which goes into context files which help interpret the data but are not dirac tags and which do include logged values over the course of the run (so the must be after).

les67: Right, the rf_roi info should only be relevant to include once per run and it's most appropriate to send it with the metadata file (that's created before the run in _do_prerun_gets) for the reasons stated by Ben. Line 155 is only populating the post-run snapshot of the slow control with the values in the endpoint_{numeric/string}_data tables of the db logged during the run.

laroque:
We shouldn't spend too much time worrying about the debug mode. That was
created before insectarium existed so that daq could be developed /tested
without needing to duplicate the rest. Insectarium takes care of that.

PR #74, opened 2016-11-15
Feature/daq provider cleaning

les67:
- Removed debug flags from DAQProvider which are not needed now that we have a working virtual system in insectarium. The <code>determine_RF_ROI</code> method in DAQProvider now broadcasts a logger.warning message noting that this method is skipped unless initiating an object with a subclass (which as of now is only RSAAcquisitionInterface)

- We now generate one single snapshot file instead of two. This file contains the latest and the logs output of SQLSnapshot. Its content will be sent to mdreceiver at the end of run.  

- Changed some logging levels to critical in SQLSnapshot. These are related to it not being able to find an endpoint with a given name or any entries in the db between two timestamps. 

- All tested in insectarium/feature/DAQProviderCleaning and working well.


les67:
I will make the appropriate change (i.e. removing the dbug flag) in the daq service config file in hardware/develop.

Issue #27, opened 2016-02-03
run interface poor error handling


laroque:
If you try to do a run with a name that is too long for the database insert, the database service returns a useful message but the run_interface object does not pass that error info back to the user, making the situation difficult to debug. Should figure out where the error is being handled and be sure to pass the return code all the way back to the user.

les67:
#30 #46

les67:
A couple of notes: 

1. The maximum number of characters (not bytes) for run_name in the runs.run_table is 80; run_name is of type varchar(80).
2. Upon insertion of a run name with 81 characters:
```
2016-11-17T21:45:34[INFO    ] dripline.core.service(316) -> received a message
2016-11-17T21:45:34[DEBUG   ] dripline.core.service(363) -> Acknowledging message 1
2016-11-17T21:45:34[DEBUG   ] dripline.core.message(166) -> original msg was: {"timestamp": "2016-11-17T21:45:34Z", "msgop": 9, "msgtype": 3, "sender_info": {"username": "root", "exe": "/usr/local/bin/dragonfly", "package": "dripline", "service_name": "", "hostname": "6e2410a9e12f", "version": "v0.0.0", "commit": ""}, "payload": {"values": [], "run_name": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaab"}}
2016-11-17T21:45:34[DEBUG   ] dripline.core.message(139) -> from_dict is: {u'timestamp': u'2016-11-17T21:45:34Z', u'payload': {u'values': [], u'run_name': u'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaab'}, u'msgtype': 3, u'sender_info': {u'username': u'root', u'exe': u'/usr/local/bin/dragonfly', u'package': u'dripline', u'service_name': u'', u'hostname': u'6e2410a9e12f', u'version': u'v0.0.0', u'commit': u''}, u'msgop': 9}
2016-11-17T21:45:34[DEBUG   ] dripline.core.message(148) -> msgtype is 3
2016-11-17T21:45:34[INFO    ] dripline.core.spimescape(63) -> request received by run_table_service
[...]

Issue #30, opened 2016-02-03
sensor logger shouldn't crash on bad insert


laroque:
we had a miss-named endpoint (middle_current_limit instead of middle_coil_current_limit I think) that produced the following traceback. Posting to slack is good (ie that a critical message posts) but the process should _not_ have crashed. It appears that there was a python IndexError which wasn't caught (I think because the query to the endpoint_id_map returned no results on the unknown sensor name).

```
2016-02-03T10:38:29[INFO    ] dripline.core.service(315) -> received a message
2016-02-03T10:38:29[WARNING ] dripline.core.gogol(51) -> got an exception:
list index out of range
2016-02-03T10:38:29[CRITICAL] dripline.core.service(414) -> Service <slow_control_table_interface> crashing with error message:
list index out of range
supposed to do an emit
2016-02-03T10:38:29[ERROR   ] dripline.core.service(415) -> Traceback (most recent call last):
  File "/home/project8/python_environments/drip3.2_dragon1.0/local/lib/python2.7/site-packages/dripline-wp2.0.0_python3.2.3_0_g1c92492-py2.7.egg/dripline/core/service.py", line 412, in run
    self._connection.ioloop.start()
  File "/home/project8/python_environments/drip3.2_dragon1.0/local/lib/python2.7/site-packages/pika/adapters/select_connection.py", line 138, in start
    self.poller.start()
  File "/home/project8/python_environments/drip3.2_dragon1.0/local/lib/python2.7/site-packages/pika/adapters/select_connection.py", line 433, in start
    self.poll()
  File "/home/project8/python_environments/drip3.2_dragon1.0/local/lib/python2.7/site-packages/pika/adapters/select_connection.py", line 494, in poll
    self._handler(fileno, event, write_only=write_only)
  File "/home/project8/python_environments/drip3.2_dragon1.0/local/lib/python2.7/site-packages/pika/adapters/base_connection.py", line 322, in _handle_events
    self._handle_read()
  File "/home/project8/python_environments/drip3.2_dragon1.0/local/lib/python2.7/site-packages/pika/adapters/base_connection.py", line 351, in _handle_read
    self._on_data_available(data)
  File "/home/project8/python_environments/drip3.2_dragon1.0/local/lib/python2.7/site-packages/pika/connection.py", line 1285, in _on_data_available
    self._process_frame(frame_value)
[...]

PR #78, opened 2016-11-29
Bug/database issues

les67:
This update resolves issues #27 and #30 including the related issue of the sensor logger service crashing when receiving a broadcast.ping command. First, a misnamed enpoint whose value we attempt to log now broadcasts a useful logger.critical message and the service continues onto the next without crashing. Second, a run with a missing run name or a run name that is too long for the db raises an error that is now caught and the error message propagated to logger.critical and the user. In the general case that the run name is not populated then, in DAQProvider, the rest of the run procedures - that is ALL run procedures - are prevented from initializing; the service continues without crashing. Lastly, broadcast.ping is parsed correctly by the logger service and doesn't crash it. All tested in insectarium (branch of same name), and working well.

### DIFF
--- a/dragonfly/implementations/daq_run_interface.py
+++ b/dragonfly/implementations/daq_run_interface.py
@@ -114,13 +114,10 @@ class DAQProvider(core.Provider):
     def start_run(self, run_name):
         '''
         '''
-        try:
-            logger.debug('self._run_name before setting is: {}'.format(self._run_name))
-            self.run_name = run_name
-            logger.debug('self._run_name after setting is: {}'.format(self._run_name))
-        except Exception as err:
-            logger.critical('all internal run procedures will be squashed')
-            raise
+        self.run_name = run_name
+        if self._run_name is None:
+            logger.critical('run_name is not present; all internal run procedures will be squashed')
+            raise core.exceptions.DriplineValueError('<{}> instance <{}> requires a value for "{}" to initialize run and procedures'.format(self.__class__.__name__, self.name, '_run_name'))
         self._run_meta = {'DAQ': self.daq_name,
                          }
         self._run_snapshot = {'LATEST':{},'LOGS':{}}
--- a/dragonfly/implementations/daq_run_interface.py
+++ b/dragonfly/implementations/daq_run_interface.py
@@ -76,6 +76,7 @@ class DAQProvider(core.Provider):
         self._stop_handle = None
         self._run_name = None
         self.run_id = None
+        self._start_time = None
         self._acquisition_count = None
 
     @property
@@ -85,9 +86,12 @@ class DAQProvider(core.Provider):
     def run_name(self, value):
         self._run_name = value
         self._acquisition_count = 0
-        result = self.provider.cmd(self.run_table_endpoint, 'do_insert', payload={'run_name':value})
-        self.run_id = result['run_id']
-        self._start_time = result['start_timestamp']
+        try:
+            result = self.provider.cmd(self.run_table_endpoint, 'do_insert', payload={'run_name':value})
+            self.run_id = result['run_id']
+            self._start_time = result['start_timestamp']
+        except core.exceptions.DriplineDatabaseError as dripline_error:
+            logger.error('failed to insert run_name to the db and obtain run_id and start_timestamp:\n{}'.format(dripline_error.message))
 
     def end_run(self):
         self._do_postrun_gets()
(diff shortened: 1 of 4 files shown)


guiguem:
- [ ] Not sure the ROI determination is optional, saving RSA data without knowing/saving where we are looking at it bad. A RaiseException prevents us from not saving them.

dragonfly/implementations/daq_run_interface.py
```diff
@@ -128,7 +142,7 @@ def _do_postrun_gets(self):
             self._run_snapshot['LOGS'].update(these_snaps)
 
     def determine_RF_ROI(self):
-        raise core.exceptions.DriplineMethodNotSupportedError('subclass must implement RF ROI determination')
+        logger.warning('subclass must implement RF ROI determination; skipped')
```
guiguem: Not sure the ROI determination is optional, saving RSA data without knowing/saving where we are looking at it bad. A RaiseException prevents us from not saving them.

dragonfly/implementations/daq_run_interface.py
```diff
@@ -101,6 +112,9 @@ def start_run(self, run_name):
         '''
         '''
         self.run_name = run_name
+        if self._run_name is None:
+            logger.critical('run_name is not present; all internal run procedures will be squashed')
+            raise core.exceptions.DriplineValueError('<{}> instance <{}> requires a value for "{}" to initialize run and procedures'.format(self.__class__.__name__, self.name, '_run_name'))
```
guiguem: Not sure also about the pertinence of a logger.critical followed by a raise Exception

dragonfly/implementations/daq_run_interface.py
```diff
@@ -82,10 +83,20 @@ def run_name(self):
...
+            self._start_time = result['start_timestamp']
+        except Exception as err:
+            logger.critical('failed to insert run_name to the db, obtain run_id, and start_timestamp. error:\n{}'.format(str(err)))
+            # end the run
+            if self._stop_handle is not None:
+                self.service._connection.remove_timeout(self._stop_handle)
+                logger.critical('run <{}> was not started'.format(value))
+                self._stop_handle = None
+                self._run_name = None
+                self.run_id = None
```
guiguem: Resetting the _run_name to None here will trigger the error in line 116-117 which does not correspond to what is happening.
A Raise Execption to stop things right at the end of this exception might be cleaner.

les67: Ok so you think it would be better to move raising the exception from line 117 to (right) after line 98? Then lines 115-117 would be removed. That does seem cleaner so I would agree.

dragonfly/implementations/daq_run_interface.py
les67: About the logger critical, my thought was that we would want to communicate the failure of this run to the user and slack. But I agree that it doesn't fit well. An exception will be raised either when the run_name insert and return fails. Is the exception message good enough then?

dragonfly/implementations/daq_run_interface.py
les67: This specific method will only be called if your daq is an instance of DAQProvider rather than RSAAcquistionInterface (or Roach in the future). For the latter, the determine_RF_ROI method is implemented and make good sense. In a real situation, the config file won't use DAQProvider as the module for the interface so it's rf roi method won't be called and it wouldn't make sense anyways if it's called. Having a warning instead of (or anything but) raising an error makes it possible to test general DAQ implementations in insectarium as well which is the only place I would see it being used. If you think that raising an error in this case makes sense also and would want to keep it as it was, then I suppose we could add a flag to skip this method in "debug" mode.

guiguem: I think the determine_RF_ROI is required in general and should not just send a warning in production mode.
I don't see how testing the DAQInterface makes sense on insectarium from a production mode testing, since no DAQ is actually here. 
Also nothing prevent you (when testing on insectarium) from making a dummy determine RF ROI when testing on insectarium (dummy could mean an empty method or commenting the actual commands sent to get the parameters and replace the values start_freq... y dummy values).

dragonfly/implementations/daq_run_interface.py
guiguem: I guess a raise exception will trigger a logger.critical, which will then be sent to the user on slack.
May be wrong here, this would require a proper test.
But again there is a self.run_name and a self._run_name in the class...

dragonfly/implementations/daq_run_interface.py
guiguem: yes.

les67:
Updating latest comments into branch and testing. Will re-submit pull request later.

PR #89, opened 2017-01-25
Feature/generic daq set condition

guiguem:
As discussed on Slack, I have moved a lot of the items in the start_run of the RSADAQInterface into the base class DAQProvider.
This makes the code much cleaner, but also shows what the required DAQ-specific methods for every new DAQ system are (currently is_running, _do_checks and determine_RF_ROI).
This new structure allows more flexibility invading new features across all DAQ system and helps to keep them coherent.
The set_condition is now a DAQ_Provider class thing, the language used to comment in the logger is supposed to be generic. 
The set_condition also allows us to stop the current run (if is_running returns True): this could be useful if we want to decrease the limit of the disk_monitor_interface.

@laroque @nsoblath Could you have a look and check that this is flexible enough for the ROACH?


guiguem:
Noah, Christine and I discussed about these changes: this will make the ROACH integration easier and the two DAQ systems consistent.

PR #101, opened 2017-05-11
Feature/roachinterface

guiguem:
Feature/roachinterface contains r2daq, roach2interface, PsyllidProvider and new classes in daq_run_interface. 
The classes MultiPsyllidProvider and ROACH1ChAcquisitionInterface are ready for streaming data acquisition with single psyllid streams.


configs/detector_status_display_service_modified.yaml
```diff
@@ -0,0 +1,6 @@
+name: status_multigets
```
guiguem: This file should be removed.

examples/kv_store_tutorial.yaml
```diff
@@ -1,5 +1,5 @@
 name: my_store
-broker: localhost
+broker: higgsino.physics.ucsb.edu
```
guiguem: change back to localhost

requirements.txt
```diff
@@ -32,3 +32,12 @@ snowballstemmer==1.2.0
 traitlets==4.0.0
 websocket-client==0.34.0
 wsgiref==0.1.2
+
```
guiguem: move this to setup.py
@nsoblath is there a graceful way on getting the specific git commit installed? submodule maybe?

laroque: `git describe` (possibly with various formatting flags) will tell you the commit you're on, the most recently tagged ancestor, commit distance etc. Is that what you want?

guiguem: no, they need the installation of a specific commit from another git repo. 
```+-e git+git://github.com/sma-wideband/adc_tests.git@65a2ef4e1cf68bee35176a1171d923a73952e13e#egg=adc5
```

dragonfly/implementations/PsyllidProvider.py
```diff
@@ -0,0 +1,404 @@
...
+from dripline import core
+from dragonfly.implementations import EthernetProvider
+
+
+__all__ = []
+
+logger = logging.getLogger(__name__)
+
+__all__.append('PsyllidProvider')
+class PsyllidProvider(core.Provider, core.Spime):
```
wcpettus: is there a reason this inherits from core.Spime?

dragonfly/implementations/PsyllidProvider.py
```diff
@@ -0,0 +1,404 @@
+'''
```
wcpettus: naming convention: use snake_case (psyllid_provider.py), the class names can remain CamelCase

dragonfly/implementations/PsyllidProvider.py
laroque: Also please convert hard tabs to soft tabs. If you use vi(m) then `:%s/\t/    /g` will do it (but you'll need to make sure you don't break any indentation blocks).

dragonfly/implementations/PsyllidProvider.py
```diff
@@ -0,0 +1,404 @@
...
+        else: return True
+
+
+    def reactivate(self, channel):
+        if self.status_value_dict[channel] == 4:
+            logger.info('Reactivating Psyllid instance of channel {}'.format(channel))
+            result = self.provider.cmd(self.queue_dict[channel], 'reactivate-daq')
+            logger.info(result)
+        else:
+            logger.warning('Cannot not reactivate Psyllid instance of channel {}'.format(channel))
```
wcpettus: Do you mean a double-negative here?

dragonfly/implementations/PsyllidProvider.py
```diff
@@ -0,0 +1,404 @@
...
+        try:
+            if self.request_status()==0:
+                self.activate()
+        except:
+            raise core.exceptions.DriplineGenericDAQError('Psyllid instance with this queue is not responding')
+
+        self.mode = 'streaming'
+        try:
+            self.set_default_central_frequencies()
+        except:
```
laroque: Can you clarify the logic here? A try/except block that doesn't catch only specific exceptions is pretty dangerous.

dragonfly/implementations/PsyllidProvider.py
```diff
@@ -0,0 +1,404 @@
...
+        
+    def set_default_central_frequencies(self):
+        freq_dict = {'a': 800e6, 'b': 800e6, 'c': 800e6}
+        self.set_all_central_frequencies(freq_dict)
+        
+
+    def request_status(self):
+        logger.info('Checking Psyllid status')
+        
+        try:
```
laroque: same, a try/except with a blanket except is pretty dangerous. What is producing the the exception you would catch and what exception(s) do you expect? We should be able to check the return code on the get().

dragonfly/implementations/PsyllidProvider.py
```diff
@@ -0,0 +1,404 @@
...
+            result = self.provider.get(self.psyllid_queue+'.daq-status', timeout=10)
+            self.status = result['server']['status']
+            self.status_value = result['server']['status-value']
+            logger.info('Psyllid is running. Status is {}'.format(self.status))
+            logger.info('Status in numbers: {}'.format(self.status_value))
+            return self.status_value
```
laroque: What does this return status_value do? If this is a return code then it needs to be in a ReplyMessage's ret_code field, not in a payload field

dragonfly/implementations/PsyllidProvider.py
```diff
@@ -0,0 +1,404 @@
...
+
+        except:
+            logger.warning('Psyllid is not running or sth. else is wrong')
+            self.status=None
+            self.status_value=None
+            logger.info('Status is {}'.format(self.status))
+            return self.status_value
```
laroque: Is this return something send back in a Reply? Status (success/failure) information needs to be in a return code, not a payload value. If "Status" means 'State' (as in "activated" or "idle" or whatever, where 'State' is some configurable parameter then maybe this is okay, but returning it in an except block seems questionable.

dragonfly/implementations/PsyllidProvider.py
```diff
@@ -0,0 +1,404 @@
...
+    def get_number_of_streams(self):
+        active_channels = [i for i in self.freq_dict.keys() if self.freq_dict[i]!=None]
+        logger.info('Active channels are {}'.format(active_channels))
+        return len(active_channels)
+
+        
+    def get_active_streams(self):
+        active_channels = [i for i in self.freq_dict.keys() if self.freq_dict[i]!=None]
+        return active_channels
+
```
laroque: There's nothing wrong with these `get_*` methods, but a nicer solution would be something like:

```
@property
def get_<whatever_thing>(self):
    <same implementation you have above>
```

because then you would be able to do `dragonfly get psyllid_provider_name <whatever_thing>` rather than `dragonfly cmd psyllid_provider_name.get_<whatever_thing>` (assuming these methods are intended for use with dragonfly and not just internally).

dragonfly/implementations/PsyllidProvider.py
```diff
@@ -0,0 +1,404 @@
...
+
+
+__all__.append('MultiPsyllidProvider')
+class MultiPsyllidProvider(core.Provider, core.Spime):
```
laroque: This class appears to implement some methods which are the same as those above. If that's correct then we should have `class MultiPsyllidProvider(PsyllidProvider)` and only implement them once. You can then "override" the methods if the Multi- version of the class needs to use different logic, but will otherwise reuse the existing logic. It is also possible for the subclass to call the implementation from the parent but do additional logic, if that's what you need.

dragonfly/implementations/PsyllidProvider.py
wcpettus: just to nitpick, it should probably be 
```python
@property
def <whatever_thing>(self):
```
to avoid having to `dragonfly get psyllid_provider_name.get_<whatever_thing>` with extra get

dragonfly/implementations/PsyllidProvider.py
```diff
@@ -0,0 +1,404 @@
...
+                    result = self.provider.set(self.queue_dict[channel]+request, cf)
+                    stream_count += 1
+                    self.mode_dict[channel]='triggered'
+                except:
+                    pass  
+        logger.info('Number of streams for channel {}: {}'.format(channel, stream_count))
+        return stream_count
+
+
+    def set_fmt_snr_threshold(self, channel='a', snr=6):
```
laroque: As with get_* above, you can use `@<property>.setter` to make attributes which have function implementations. This is more pythonic and as above, will allow access via `dragonfly get` and `dragonfly set`.

For example, if the "tau" property is defined as 2x the "pi" property, you could do:
```
@property
def tau(self):
    return self.pi*2.
@tau.setter
def tau(self, value)
    self.pi = value/2.
```

dragonfly/implementations/PsyllidProvider.py
```diff
@@ -0,0 +1,404 @@
...
+        result = self.provider.cmd(self.queue_dict[channel], request, payload=payload)
+
+        logger.info('back to time and freq')
+        request = 'run-daq-cmd.ch'+str(self.channel_dict[channel])+'.tfrr.time-and-freq'
+        result = self.provider.cmd(self.queue_dict[channel],request)
+
+        logger.info('apply-trigger')
+        request = 'run-daq-cmd.ch'+str(self.channel_dict[channel])+'.fmt.apply-trigger'
+        result = self.provider.cmd(self.queue_dict[channel],request)
+
```
laroque: There are lots of calls to self.provider.[get,set,cmd]. Are the return codes being checked on all of these calls? If there is an error somewhere, will a useful return code eventually make it back to the user?

dragonfly/implementations/PsyllidProvider.py
laroque: fixed in response to wcpettus, that is what I intended

dragonfly/implementations/PsyllidProvider.py
wcpettus: `provider.[get,set,cmd]` uses `provider._send_request`, which has a built in check for error codes + raise exception; as long as these aren't embedded in try-except loops won't that be sufficient?

dragonfly/implementations/daq_run_interface.py
```diff
@@ -26,6 +29,7 @@ def __init__(self,
                  directory_path=None,
                  data_directory_path=None,
                  meta_data_directory_path=None,
+                 #Linux_not_Windows_DAQ = True,
```
laroque: If this isn't used please delete it and the commented lines below that were based on it.

dragonfly/implementations/PsyllidProvider.py
cclaessens: If you don't know a reason then there probably is none. When I wrote that I looked at what some of the other classes inherited from at the time.

dragonfly/implementations/daq_run_interface.py
```diff
@@ -332,4 +343,613 @@ def save_trace(self, trace, comment):
             logger.debug("things are about to be dumped in file")
             json.dump(result_meta, outfile, indent=4)
             logger.debug("things have been dumped in file")
-        logger.info("saving {}: successful".format(path))
+            logger.info("saving {}: successful".format(path))
+
+
+
+__all__.append('ROACH1ChAcquisitionInterface')
```
laroque: Let's split this class into its own file, it is probably fine for the single channel and multi-channel roach classes to be in the same file. Also, if there area methods in common between the two then perhaps one could inherit from the other, or there should be a base class that they both extend (I haven't looked through the details enough to be sure if that makes sense).

dragonfly/implementations/PsyllidProvider.py
cclaessens: These get and set methods are only accessed internally and never by the operator. Do you still want them to be changed?

dragonfly/implementations/daq_run_interface.py
```diff
@@ -332,4 +343,613 @@ def save_trace(self, trace, comment):
...
+                        payload = {'channel':self.channel_id, 'filename': os.path.join(directory, psyllid_filename), 'duration':duration}
+                        result = self.provider.cmd(self.psyllid_interface, 'start_run', payload=payload)
+                    # acquisitions with duration = max_duration
+                    else:
+                        total_run_time+=max_duration
+                        logger.info('Going to tell psyllid to start run')
+                        payload = {'channel':self.channel_id, 'filename': os.path.join(directory, psyllid_filename), 'duration':max_duration}
+                        result = self.provider.cmd(self.psyllid_interface, 'start_run', payload=payload)
+                    time.sleep(self._max_duration)
+        except:
```
laroque: same try/except comment from above

dragonfly/implementations/daq_run_interface.py
```diff
@@ -332,4 +343,613 @@ def save_trace(self, trace, comment):
...
+        result = self.provider.cmd(self.psyllid_interface, 'make_trigger_mask', payload=payload)
+        if result['values'][0]!=True:
+            return False
+        else: return True
+
+
+
+
+__all__.append('ROACHMultiChAcquisitionInterface')
+class ROACHMultiChAcquisitionInterface(DAQProvider, core.Spime):
```
laroque: Same questions, is this really a Spime? can it go with the other ROACH class into its own source file? should this inherit from the single channel version, or should they share a common (possibly abstract) base class?

dragonfly/implementations/r2daq.py
```diff
@@ -0,0 +1,1442 @@
...
+
+from __future__ import division
+
+import adc5g
+from copy import deepcopy
+from corr.katcp_wrapper import FpgaClient
+from datetime import datetime
+import logging
+import netifaces as ni
+from numpy import (
```
laroque: This is really a style/personal preference thing, but I don't like importing lots of elements `from` a library, especially something like `abs` which is replacing an existing core-python function. Instead it is nice to either use `numpy.abs` etc (so it is clear which version of abs or whatever you are using), or you can shorten it if you like with `import numpy as np` and `np.abs`.

dragonfly/implementations/PsyllidProvider.py
wcpettus: if you make it an @property, you can still access it with `getattr(self, "<whatever_thing>")`, but it has a cleaner implementation in case we ever access it from outside

similarly an @<whatever_thing>.setter can be set with `setattr(self, "<whatever_thing>", val)` wherever you use that

dragonfly/implementations/roach2interface.py
```diff
@@ -0,0 +1,429 @@
...
+__all__.append('Roach2Provider')
+class Roach2Provider(ArtooDaq, core.Provider):
+    '''
+    A DAQProvider for interacting with the Roach
+    '''
+    def __init__(self, **kwargs):
+
+
+        core.Provider.__init__(self, **kwargs)
+
```
laroque: do we not need to `ArtooDaq.__init__()` here also?

dragonfly/implementations/PsyllidProvider.py
```diff
@@ -0,0 +1,404 @@
...
+import logging
+import uuid
+import time
+import os
+import json
+from datetime import datetime
+
+# internal imports
+from dripline import core
+from dragonfly.implementations import EthernetProvider
```
wcpettus: check your imports - EthernetProvider, uuid, os, and json all aren't called

dragonfly/implementations/roach2interface.py
```diff
@@ -0,0 +1,429 @@
...
+
+
+__all__.append('Roach2Interface')
+class Roach2Interface(Roach2Provider, EthernetProvider):
+    def __init__(self,
+                 roach2_hostname = 'led',
+                 source_ip_a = None,
```
laroque: Sorry I missed this, it would probably have been much cleaner to use a dictionary of configurations for each psyllid target, something like:

```python
psyllid_a_config = {'source_ip': None,
                               ' source_port': None,
                               ...,
                              }
```
And then have the class store all of those in a dictionary too:
```python
psyllid_configs{'a': psyllid_a_config,
                        'b': psyllid_b_config,
                         ...
                        }
```
That way you could take all the places where you treat each channel and instead have one block of code that loops over the channels and uses the same code with `psyllid_configs['a']['source_ip'] etc.

dragonfly/implementations/roach2interface.py
```diff
@@ -0,0 +1,429 @@
...
+    '''
+    def __init__(self, **kwargs):
+
+
+        core.Provider.__init__(self, **kwargs)
+
+
+
+__all__.append('Roach2Interface')
+class Roach2Interface(Roach2Provider, EthernetProvider):
```
wcpettus: can remove EthernetProvider inheritance

dragonfly/implementations/PsyllidProvider.py
cclaessens: The idea is that if request_status fails to get a response from psyllid there is no psyllid instance running or the queue is wrong. In any case you cannot use it for data acquisition. 
But request status also catches an error so I dont see how this is working right now...
The PsyllidProvider class as opposed to the MultiPsyllidProvider is not intended to be used for the single channel acquisition and needs more changes and testing anyway.
But you re right the exceptions should all be specific.

dragonfly/implementations/PsyllidProvider.py
cclaessens: Status means state and the number will tell the daq run interface how to proceed.

dragonfly/implementations/PsyllidProvider.py
cclaessens: No :)

dragonfly/implementations/roach2interface.py
cclaessens: no, only in the setup call after making a dictionary with the ports aso which is passed to ArtooDaq.__init__()

dragonfly/implementations/roach2interface.py
cclaessens: Historically we started with one channel only and had a few parameters only. After I added the other channels and the list of parameters became very long I thought I should turn it into a dictionary at some point but never did...

dragonfly/implementations/daq_run_interface.py
cclaessens: I would like to make a new feature for the multi channel classes and remove them from this branch. I havent worked on them since march and they are not ready for usage.
I would suggest to leave the single channel classes where they are right now and when the multi channel classes will be added to develop one day we can move them all into a new source file.

requirements.txt
nsoblath: Do you mean a way of ensuring that the sma-wideband/adc_tests git repo is checked out to a particular commit?  You can just run `git checkout [commit]`, I think.

wcpettus: The issue is that the current installation method deprecates use of the requirements.txt in favor of extras_require in setup.py.  Using something like this should accomplish this (not sure how it differs between a core and an extra requirement):
https://mike.zwobble.org/2013/05/adding-git-or-hg-or-svn-dependencies-in-setup-py/

dragonfly/implementations/daq_run_interface.py
```diff
@@ -502,50 +491,67 @@ def _start_data_taking(self, directory, filename):
...
+        if NAcquisitions<=1:
+            psyllid_filename = filename+'.egg'
+            if not os.path.exists(directory):
+                os.makedirs(directory)
 
+            logger.info('Going to tell psyllid to start the run')
+            payload = {'channel':self.channel_id, 'filename': os.path.join(directory, psyllid_filename), 'duration':duration}
+            try:
+                result = self.provider.cmd(self.psyllid_interface, 'start_run', payload=payload)
+            except:
```
laroque: We still need to be explicit about what to catch in any except block. For example, if you want to catch any DriplineException that is raised during the `self.provider.cmd` you could do
```python
try:
    result = <all that stuff>
except exceptions.DriplineException as e:
    <whatever logic you need>
```

dragonfly/implementations/daq_run_interface.py
```diff
@@ -502,50 +491,67 @@ def _start_data_taking(self, directory, filename):
...
+                logger.error('Starting Psyllid run failed')
+                logger.info('unblock channel')
+                payload = {'channel': self.channel_id}
+                result = self.provider.cmd(self.daq_target, 'unblock_channel', payload=payload)
```
laroque: If `daq_target` is unavailable for some reason (maybe it crashed), then this command will also raise an exception, which will prevent the `DriplineGenericDAQError` below from ever being raised. Is that desirable? If not you should nest another try/except block that attempts to do this cleanup. The logic about what to catch is potentially reversed here, you may actually want to catch any exception and instead raise the one on the next line.

cclaessens: Do you think that is necessary? If daq_taget is unavailable the checks before the run and the blocking of the channel will not be successful. You could be unlucky and things break between these requests. Then catching anything first would be safer.

dragonfly/implementations/daq_run_interface.py
```diff
@@ -502,50 +491,67 @@ def _start_data_taking(self, directory, filename):
...
+                    total_run_time+=max_duration
+                    logger.info('Going to tell psyllid to start run')
+                    payload = {'channel':self.channel_id, 'filename': os.path.join(directory, psyllid_filename), 'duration':max_duration}
+                    try:
                         result = self.provider.cmd(self.psyllid_interface, 'start_run', payload=payload)
-                    time.sleep(self._max_duration)
-        except:
-            logger.error('Starting Psyllid run failed')
-
+                    except:
```
laroque: same exception handling comment, there is a big difference between "`psyllid_interface` returned an error" and "the local function call failed". We want to be sure not to hide the latter sort of errors.

dragonfly/implementations/daq_run_interface.py
```diff
@@ -502,50 +491,67 @@ def _start_data_taking(self, directory, filename):
...
+                        logger.error('Starting Psyllid run failed')
+                        logger.info('unblock channel')
+                        payload = {'channel': self.channel_id}
+                        result = self.provider.cmd(self.daq_target, 'unblock_channel', payload=payload)
```
laroque: again, this line could cause the following line to be missed which is probably not what we want

dragonfly/implementations/psyllid_provider.py
```diff
@@ -218,18 +68,18 @@ def request_status(self, channel):
         logger.info('Checking Psyllid status of channel {}'.format(channel))
         try:
             result = self.provider.get(self.queue_dict[channel]+'.daq-status', timeout=10)
-            logger.info(result)
+        except:
+            logger.warning('Psyllid instance for channel {} is not running or sth. else is wrong'.format(channel))
+            self.status_dict[channel]=None
+            self.status_value_dict[channel]=None
+            return self.status_value_dict[channel]
+        else:
```
laroque: The except should be qualified, after that you should consider if you want a `finally` statement to do some cleanup (see  https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions) before that exception is raised

dragonfly/implementations/psyllid_provider.py
```diff
@@ -283,69 +133,86 @@ def quit_psyllid(self, channel):
...
             elif self.mode_dict[channel] == 'triggered':
-                request = 'active-config.ch'+str(self.channel_dict[channel])+'.trw.center-freq'
+                request = 'active-config.ch'+self.channel_dict[channel]+'.trw.center-freq'
                 result = self.provider.set(self.queue_dict[channel]+'.'+request, cf)
                 logger.info('Set central frequency of egg writer for channel {} to {} Hz'.format(channel, cf))
                 self.freq_dict[channel]=cf
                 return True
+            else:
+                return False
         except:
```
laroque: same exception handling comments

dragonfly/implementations/daq_run_interface.py
```diff
@@ -332,4 +338,324 @@ def save_trace(self, trace, comment):
...
+class ROACH1ChAcquisitionInterface(DAQProvider):
+    '''
+    A DAQProvider for interacting with ROACH-Psyllid DAQ
+    '''
+    def __init__(self,
+                 psyllid_interface='psyllid_interface',
+                 daq_target = 'roach2_interface',
+                 channel = 'a',
+                 filename_prefix = 'psyllid',
+                 hf_lo_freq = None,
```
wcpettus: Shouldn't `hf_lo_freq` be a global parameter, since it is not DAQ specific (unless we ever plan to digitize off a different mixing or at raw frequency?).  This is for @laroque probably and is not ROACH-specific.

dragonfly/implementations/daq_run_interface.py
```diff
@@ -332,4 +338,324 @@ def save_trace(self, trace, comment):
...
+__all__.append('ROACH1ChAcquisitionInterface')
+class ROACH1ChAcquisitionInterface(DAQProvider):
+    '''
+    A DAQProvider for interacting with ROACH-Psyllid DAQ
+    '''
+    def __init__(self,
+                 psyllid_interface='psyllid_interface',
+                 daq_target = 'roach2_interface',
+                 channel = 'a',
+                 filename_prefix = 'psyllid',
```
wcpettus: remove `filename_prefix`, as it is a defined variable in DAQProvider (unless you really want to overwrite the default value)

dragonfly/implementations/daq_run_interface.py
```diff
@@ -332,4 +338,324 @@ def save_trace(self, trace, comment):
...
+
+
+__all__.append('ROACH1ChAcquisitionInterface')
+class ROACH1ChAcquisitionInterface(DAQProvider):
+    '''
+    A DAQProvider for interacting with ROACH-Psyllid DAQ
+    '''
+    def __init__(self,
+                 psyllid_interface='psyllid_interface',
+                 daq_target = 'roach2_interface',
```
wcpettus: `daq_target`could also be a global parameter of DAQProvider?

dragonfly/implementations/daq_run_interface.py
```diff
@@ -332,4 +338,324 @@ def save_trace(self, trace, comment):
...
+        self.status_value = None
+        self.channel_id = channel
+        self.freq_dict = {self.channel_id: None}
+        self._max_duration = 0
+        self._run_time = 1
+        self._run_name = "test"
+        self.payload_channel = {'channel':self.channel_id}
+
+        if hf_lo_freq is None:
+            raise core.exceptions.DriplineValueError('the Psyllid acquisition interface requires a "hf_lo_freq" in its config file')
```
wcpettus: error message should reflect this is "ROACH2 acquisition..."

wcpettus:
prologix_provider.py could be removed from the pull request, it just has a permissions change from 644 to 755 which is unnecessary

dragonfly/implementations/roach2interface.py
```diff
@@ -0,0 +1,316 @@
+# -*- coding: utf-8 -*-
```
wcpettus: snake_case convention for file names: this should be `roach2_interface.py`

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -0,0 +1,338 @@
...
+
+        self._run_meta['RF_ROI_MIN'] = float(cf-50e6) + float(self._hf_lo_freq)
+        logger.debug('RF Min: {}'.format(self._run_meta['RF_ROI_MIN']))
+
+        self._run_meta['RF_ROI_MAX'] = float(cf+50e6) + float(self._hf_lo_freq)
+        logger.debug('RF Max: {}'.format(self._run_meta['RF_ROI_MAX']))
+
+
+
+    def _start_data_taking(self, directory, filename):
```
wcpettus: the roach DAQ should not time the run with a `sleep`, it should use an `add_timeout` like the RSA and then have the appropriate `remove_timeout` in the stop method

```python
    def _start_data_taking(self,directory,filename):
        self.provider.cmd(self._daq_target, 'start_run', [directory, filename])
        logger.info("Adding {} sec timeout for run <{}> duration".format(self._run_time, self.run_id))
        self._stop_handle = self.service._connection.add_timeout(self._run_time, self.end_run)

    def _stop_data_taking(self):
        self.provider.cmd(self._daq_target, 'end_run')
        if self._stop_handle is not None:
            logger.info("Removing sec timeout for run <{}> duration".format(self.run_id))
            self.service._connection.remove_timeout(self._stop_handle)
            self._stop_handle = None
```
(why is this in the RSA and not the general methods @guiguem?)

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -0,0 +1,338 @@
...
+                        payload = {'channel': self.channel_id}
+                        try:
+                            result = self.provider.cmd(self.daq_target, 'unblock_channel', payload=payload)
+                        finally:
+                            raise e
+                time.sleep(self._max_duration)
+
+        logger.info('unblock channel')
+        payload = {'channel': self.channel_id}
+        result = self.provider.cmd(self.daq_target, 'unblock_channel', payload=payload)
```
wcpettus: this should be in the stop method?

dragonfly/implementations/__init__.py
```diff
@@ -28,3 +28,9 @@
 # keep these out of sequence, they inherit from elsewhere in dragonfly
 from .lockin_provider import *
 from .muxer_provider import *
+
+#from .sql_snapshot import *
```
laroque: do these imports need to be out of sequence for some reason? If so please describe in a comment line, otherwise please alphabetize them with the imports above.

dragonfly/implementations/psyllid_provider.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def request_status(self, channel):
+        logger.info('Checking Psyllid status of channel {}'.format(channel))
+        try:
+            result = self.provider.get(self.queue_dict[channel]+'.daq-status', timeout=10)
+        except core.exceptions.DriplineError:
+            logger.warning('Psyllid instance for channel {} is not running or returned error'.format(channel))
+            self.status_dict[channel]=None
+            self.status_value_dict[channel]=None
+            return self.status_value_dict[channel]
```
laroque: Sorry, I know this form exists in several places but I think I understand the logic better now. Please correct the following:

1) attempt to send a Request to the psyllid service asking for the `daq-status`
2.1) If the a normal Reply is received, do the stuff in the `else` block. (A couple of questions here: i) is there a reason to put it in an else block rather than as direct statements at one less indentation level? I assume not because every exception you catch either returns or raises a new exception, but this perhaps more explicit so I'm not suggesting changing it. ii) Is the purposes of self.status_dict and status_value_dict simply to cache the state of psyllid services, and if so is there ever a case where you care about those states when you use the cached value rather than polling again? My guess is that you don't actually need to cache the value (it is more consistent dripline design not to do so), and that at most you'd want to save those states in a local variable for things like getting/checking the state of all the required psyllid instances before starting a run that talks to several. But I haven't gone through the logic of the full file so maybe I'm missing something
2.2) If the reply failed you set the cache values to None and then return that. I wonder if this is really the desired behavior. Is it ever the case that this happens and it is not an indication that the user needs to go start/restart psyllid or otherwise check it for errors? I assume we want to pass something back to the user so that they know that. Either by re-raising the existing exception, or raising a new one with more information if we have it. If there are a mix of cases we could also return a warning returncode (value less than 100) which is allowed to include a payload with useful data for the parent process to continue executing, but also indicates that something went wrong-ish.

dragonfly/implementations/psyllid_provider.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def activate(self, channel):
+        if self.status_value_dict[channel] == 0:
+            logger.info('Activating Psyllid instance for channel {}'.format(channel))
+            result = self.provider.cmd(self.queue_dict[channel], 'activate-daq')
+        else:
+            logger.warning('Cannot activate Psyllid instance of channel {}'.format(channel))
+            return False
+        time.sleep(1)
+        self.request_status(channel)
```
laroque: Are there cases where the "activate-daq" command returns a success code but requesting the status does not return 4? If not then the return code on the provider.cmd() call probably covers this block. If yes then it is probably useful to return more detailed information. For example, if activate fails, shouldn't that be an exception here?

dragonfly/implementations/psyllid_provider.py
```diff
@@ -0,0 +1,279 @@
...
+        if self.status_value_dict[channel]!=4:
+            logger.warning('Activating failed')
+            return False
+        else: return True
```
laroque: Just a styling note. I always avoid same-line logic blocks, ie I'd do:
```
else:
    return True
```
not
```
else: return True
```

The reasoning being that it keeps everything uniform everywhere, and also that it means if you want to add something (like a logger.info statement) to that code block, you only need to add a new line, you don't also have to modify the existing logic lines (which then show up as changed on a git diff, even though the logic is unchanged)

dragonfly/implementations/psyllid_provider.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+    def deactivate(self, channel):
```
laroque: both comments regarding activate apply to this method also

dragonfly/implementations/psyllid_provider.py
```diff
@@ -0,0 +1,279 @@
...
+    def reactivate(self, channel):
+        if self.status_value_dict[channel] == 4:
+            logger.info('Reactivating Psyllid instance of channel {}'.format(channel))
+            result = self.provider.cmd(self.queue_dict[channel], 'reactivate-daq')
+            logger.info(result)
+        else:
+            logger.warning('Cannot reactivate Psyllid instance of channel {}'.format(channel))
+            return False
+        time.sleep(1)
+        self.request_status(channel)
```
laroque: same questions from activate, note that you did the else block in the preferred way here

dragonfly/implementations/psyllid_provider.py
```diff
@@ -0,0 +1,279 @@
...
+            result = self.provider.get(self.queue_dict[channel]+'.'+request)
+            logger.info('Psyllid says cf is {}'.format(result['center-freq']))
+            self.freq_dict[channel]=result['center-freq']
+            return self.freq_dict[channel]
+        if self.mode_dict[channnel] == 'triggered':
+            request = 'active-config.'+self.channel_dict[channel]+'.trw'
+            result = self.provider.get(self.queue_dict[channel]+'.'+request)
+            logger.info('Psyllid says cf is {}'.format(result['center-freq']))
+            self.freq_dict[channel]=result['center-freq']
+            return self.freq_dict[channel]
```
laroque: I suggest the following rework to simplify this method (note how the repeated lines are eliminated):
```
def get_central_frequency(self, channel):
    routing_key_map = {
                                     'streaming': '.strw',
                                     'triggered': '.trw',
                                    }
    request = 'active-config.' + self.channel_dict[channel] + routing_key_map[self.mode_dict[channel]]
    result = self.provider.get(self.queue_dict[channel]+'.'+request)
    logger.info('Psyllid says cf is {}'.format(result['center-freq']))
￼    self.freq_dict[channel]=result['center-freq']
￼    return self.freq_dict[channel]
```

dragonfly/implementations/psyllid_provider.py
```diff
@@ -0,0 +1,279 @@
...
+
+    
+
+    def set_central_frequency(self, channel, cf):
```
laroque: consider a rework as in get central frequency above so that the repeated lines can be combined.

dragonfly/implementations/psyllid_provider.py
```diff
@@ -0,0 +1,279 @@
...
+        return stream_count
+
+
+    def set_fmt_snr_threshold(self, snr, channel='a'):
+        if self.mode_dict[channel] == 'streaming':
+            logger.warning('Psyllid not in streaming mode')
+            return False
+        request = '.node-config.ch'+str(self.channel_dict[channel])+'.fmt.threshold-power-snr'
+        result = self.provider.set(self.queue_dict[channel]+request, snr)
+        self.deactivate(channel)
```
laroque: Are these three lines just to prevent talking to psyllid while it is drawing a new frequency mask? If so @nsoblath could psyllid return an error code if it gets a request while drawing the mask and/or be polled to ask if it is still busy? These lines feel very clunky (what if it is much faster than 1 second and we're needlessly blocking, or in the future if it takes 1.1 seconds for example) but I could be missing something important about the roach or psyllid.

dragonfly/implementations/psyllid_provider.py
```diff
@@ -0,0 +1,279 @@
...
+            self.activate(channel)
+
+        logger.info('freq-only')
+        request = 'run-daq-cmd.ch'+str(self.channel_dict[channel])+'.tfrr.freq-only'
+        result = self.provider.cmd(self.queue_dict[channel],request)
+
+        logger.info('update-mask')
+        request = 'run-daq-cmd.ch'+str(self.channel_dict[channel])+'.fmt.update-mask'
+        result = self.provider.cmd(self.queue_dict[channel],request)
+        time.sleep(1)
```
laroque: Again, these sleeps feel very clunky. Hopefully it is the case that either: 1) the action is quick and the target is done when it sends its reply, or 2) the success Reply  only indicates that the request was received and valid but the action is ongoing and there is therefore a way to poll to see when it is done.

One of the bigger concerns here is that if we're in case 2 and not actually polling, then how do we know if there was an error in the last step of the "tfrr.freq-only" or "update-mask" commands? It isn't clear that they've had an opportunity to send a reply indicating an issue.

dragonfly/implementations/psyllid_provider.py
```diff
@@ -0,0 +1,279 @@
...
+        result = self.provider.cmd(self.queue_dict[channel], request, payload=payload)
+
+        logger.info('back to time and freq')
+        request = 'run-daq-cmd.ch'+str(self.channel_dict[channel])+'.tfrr.time-and-freq'
+        result = self.provider.cmd(self.queue_dict[channel],request)
+
+        logger.info('apply-trigger')
+        request = 'run-daq-cmd.ch'+str(self.channel_dict[channel])+'.fmt.apply-trigger'
+        result = self.provider.cmd(self.queue_dict[channel],request)
+        return True
```
laroque: One last comment that applies throughout this file. You capture the result of every provider.cmd in a local variable named `result`. I don't know that the performance hit would ever be noticed, but if you aren't going to then use the content of that variable for anything (usually doing something with the payload values) then there's no need to store it. You can just call the methods and allow the returned value to be garbage collected immediately.

dragonfly/implementations/roach2interface.py
```diff
@@ -0,0 +1,316 @@
...
+            
+
+    @property
+    def gain(self):
+        return self.gain_dict
+
+
+    @gain.setter
+    def gain(self, val):
+        gain, channel= val
```
laroque: Here you support a setter that takes a 2 element object with a gain value and channel. Earlier there was not setter for these objects and one needed to call `set_<parameter>(self, channel, value)`. I think I'm okay with either convention but I feel that it is important that it be uniform. It is also really important that the order of the unpacking be uniform throughout (`gain,ch = val` vs `ch,gain = val`) and should probably be in a doc string on these methods (I'd generally like to have more doc strings in all of the source, but it is really important when there are arguments which could be non-intuitive like those which need to be unpacked).

Less important, but I think that the reverse ordering is more intuitive (where the setting is going then the value to set). Others can disagree on that and the critical thing is only that it be consistent within the class.

dragonfly/implementations/roach2interface.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+    @property
+    def fft_shift_vector(self):
+        return self.fft_shift
+
+
+    @fft_shift_vector.setter
+    def fft_shift_vector(self, val):
+        shift, tag = val
```
laroque: This is consistent with above, please add a doc string. As above I like the setter but wish it was consistent with the previous earlier source file.

dragonfly/implementations/roach2interface.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+
+    @property
+    def roach2_clock(self):
+        board_clock = self.roach2.est_brd_clk()
+        return board_clock
+
+
+    def get_T_packets(self,dsoc_desc=None, channel='a', NPackets=10, mean=True, path=None):
```
laroque: Is this method used outside of debugging? It seems pretty strange to me that an intermediate service would be saving files to disk. I'd expect the method to be called using a dragonfly request and the result send in a reply message and then the user can write the result to a file if desired.

If this is only used for debugging then I'd say leave it but add a doc string indicating that the method is not generally intended for use. If we will use it then we should discuss the behavior further.

dragonfly/implementations/roach2interface.py
```diff
@@ -0,0 +1,316 @@
...
+        NPackets = np.shape(p)[0] 
+        if mean == True:
+            p = np.mean(p, axis = 0)
+        if path==None:
+            path=self.monitor_target
+        np.save(path+'/T_packets_ft_channel'+channel+'_cf_'+str(cf)+'Hz_gain_'+str(gain)+'_N_'+str(NPackets), p)
+        np.save(path+'/T_packets_channel'+channel+'_cf_'+str(cf)+'Hz_gain_'+str(gain)+'_N_'+str(NPackets), t)
+
+
+    def get_F_packets(self,dsoc_desc=None, channel='a', NPackets=100, mean=True):
```
laroque: same comment as for get_T_packets

dragonfly/implementations/roach2interface.py
```diff
@@ -0,0 +1,316 @@
...
+            if pkts[i].freq_not_time==True:
+                f=pkts[i].interpret_data()
+                p.append(np.abs(f))
+        p = np.array(p)
+        if mean == True:
+            p = np.mean(p, axis = 0)
+        np.save(self.monitor_target+'/F_packets_channel_'+channel+'_cf_'+str(cf)+'Hz_gain_'+str(gain), p)
+
+
+    def get_raw_adc_data(self, count=0, N=1, mean = True):
```
laroque: Same comments as getting T packets from above.

dragonfly/implementations/roach2interface.py
```diff
@@ -0,0 +1,316 @@
...
+                p.append(np.abs(np.fft.fftshift(np.fft.fft(x_all[i*16384:(i+1)*16384])))/16384)
+        p = np.array(p)
+        if mean == True:
+            p = np.mean(p, axis = 0)
+        logger.info('shape of saved array is: {}'.format(np.shape(p)))
+        np.save(self.monitor_target+'/raw_adc'+str(count), p)
+        logger.info('Raw data saved to {}'.format(self.monitor_target+'/raw_adc'+str(count)+'.npy'))
+
+
+    def calibrate_manually(self, gain1=0.0, gain2=0.42, gain3=0.42, gain4=1.55, off1=3.14, off2=-0.39, off3=2.75, off4=-1.18):
```
laroque: Could you add a comment/docstring on these initial values? Are they values from Andre or did you determine them in some way (and if so how?)? etc.

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -0,0 +1,338 @@
...
+    @property
+    def max_duration(self):
+        return self._max_duration
+
+    @max_duration.setter
+    def max_duration(self, duration):
+        self._max_duration = duration
+
+
+    def _check_roach2_status(self):
```
laroque: this method behavior is somewhat unintuitive. I'd expect "check_status" to either i) return the status, or ii) return either nothing or raise an exception if the status isn't good. It appears that the check is actually if the roach is running, could the method be `def _check_roach2_is_running(self)` instead?

I don't think I missed other cases of this, but the comment should apply elsewhere if I missed it.

dragonfly/implementations/roach_daq_run_interface.py
laroque: This is probably historic from before Mathieu started working on it. The mantis interface was done first and did not use timeouts. You'd add run requests of fixed length to the run queue then poll the queue to see when it was done/empty. Adding support for the RSA required a change because it need to be explicitly told to start/stop.

The better solution (discussed on slack) would be for me to clean up the scheduler class and for that to be used rather than dealing with pika's connection object timeouts in other classes.

dragonfly/implementations/__init__.py
wcpettus: roach2interface inherits from r2daq, roach_daq_run_interface inherits from daq_run_interface; both of these will be satisfied by straight alphabetic ordering

dragonfly/implementations/psyllid_provider.py
cclaessens: 2.1) 
i) The reason for the else block is just being explicit.
ii) whenever the status of psyllid needs to be checked it is re-requested. 
The reason why the status can be none instead of raising an error in psyllid provider is that the psyllid provider class works with up to 3 psyllid instances. If you only start one psyllid instance for example the status for the other 2 is none because they are not there. This should not prevent psyllid_provider working normally with the one psyllid instance which is there.
The roach_daq_run_interface for a single channel requests the psyllid status for this channel via the psyllid_provider. if the status is none an error will be raised in the roach_daq_run_interface.

dragonfly/implementations/psyllid_provider.py
cclaessens: Psyllid used to have problems with changing the daq-status and not getting stuck somewhere. Then activating did not result in an error and still the status could be wrong afterwards.
Currently this is not the case anymore but I still don't feel totally comfortable with just assuming the correct status after requesting a change as long as psyllid is still under development.
If activating psyllid fails and results in an error the error message from psyllid is forwarded to the operator (will double check). So I don't think we need an exception here.

dragonfly/implementations/psyllid_provider.py
cclaessens: This should just be self.reactivate(channel). The last time I tested this I got into trouble with reactivating psyllid (psyllid/sepfile) and changed it to this. It was a quick test though and more testing of the psyllid mask writing and triggering is on my list. I intended to make it an issue then if the problem persists.

I could also remove theses methods for a later merge, as this one is not intended for usage for triggered acquisition.

dragonfly/implementations/psyllid_provider.py
cclaessens: When I tested it last time the commands were sent to quickly and psyllid hadn't finished the last request and doing the next resulted in an error.

dragonfly/implementations/roach2interface.py
cclaessens: Agreed. I think they are useful enough to use them outside of debugging and they should just write the packet content to a json file.

dragonfly/implementations/roach2interface.py
cclaessens: I got these values from André and he found them using the automatic calibration method.
When the roach gets a signal from a signal generator the automatic adc calibration fails so I tried to use this values to remove the 800MHz spur. 
Currently we don't need this method but I think it might be useful one day to have more control over the adc settings.

dragonfly/implementations/roach_daq_run_interface.py
cclaessens: I changed it to check_roach2_is_ready because it will not only return false if the roach is switched off but also if the adcs have not been calibrated. The logic for that is in the roach2_interface.

dragonfly/implementations/psyllid_provider.py
```diff
@@ -0,0 +1,274 @@
...
+            logger.error('Unexpected error in set_central_frequency')
+            raise 
+
+
+    def start_run(self, channel, duration, filename):
+        payload = {'duration':duration, 'filename':filename}
+        self.provider.cmd(self.queue_dict[channel], 'start-run', payload=payload)
+
+
+    def stop_run(self, channel):
```
wcpettus: we should match naming between the RSA and ROACH/Psyllid - use `end_run`

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -0,0 +1,340 @@
...
+                        try:
+                            self.provider.cmd(self.daq_target, 'unblock_channel', payload=payload)
+                        finally:
+                            raise e
+
+
+
+    def _stop_data_taking(self):
+        if self.is_running:
+            self.provider.cmd(self.psyllid_interface, 'stop_run', payload = self.payload_channel)
```
wcpettus: same as above, name this `end_run`

dragonfly/implementations/roach_daq_run_interface.py
wcpettus: Is this functionality adequately provided in the new scheduler?

requirements_no_sql.txt
```diff
@@ -31,3 +31,6 @@ snowballstemmer==1.2.0
 traitlets==4.0.0
 websocket-client==0.34.0
 wsgiref==0.1.2
+corr==0.7.3
+-e git+git://github.com/sma-wideband/adc_tests.git@65a2ef4e1cf68bee35176a1171d923a73952e13e#egg=adc5
```
guiguem: this is solved: the adc_tests package comes from project8/adc_tests (forked from sma-wideband), and has a version control.
A tag (v0.0.1) has been created for version that currently works 65a2ef4e1cf68bee35176a1171d923a73952e13e: the associated version defined in the setup.py is 0.0.1.
In order to prevent confusion, I bumped the latest version in the master branch to 0.0.2 in the setup.py and tagged it. Installing this version with the requirements adc5g==0.0.1 in dragonfly setup.py file, will cause a crash of the installation (recommended...)

guiguem:
Resolved the latest conflicts

requirements_no_sql.txt
wcpettus: I found a slightly cleaner implementation which installs off the CfA repo and uses the "dependency_links" option.  The install command is now:
`pip install --process-dependency-links ~/test/dragonfly[colorlog,roach]`

On the install, it will claim that that "DEPRECATION: Dependency Links processing has been deprecated and will be removed in a future release." which the internet says is not exactly true.  There is no plan to truly deprecate it because they haven't replaced the functionality yet, but developers want to deprecate it because it has potential security risks.

Tested on zeppelin and confirmed to work.

dragonfly/implementations/roach_daq_run_interface.py
cclaessens: there is already an end_run method in daq provider. This method calls the _stop_data_taking method of the daq target.

dragonfly/implementations/psyllid_provider.py
cclaessens: this method only needs to be called for interrupting a run. I would like to keep this name as long as the command to send to psyllid is stop run as well.

PR #102, opened 2017-05-18
Slack channel/status message full support

guiguem:
Allow the definition of the Slack channel where to send the logger.critical: can be defined in the CLI via (-s; —slack_channel) or in the config file at the same level as the service.
The default value is 'p8_alerts', hardcoded in dragonfly and amqphandler (might be optional)...
All options have been tested with higgsino.

I generically called this `Slack channel`, but fundamentally it sends an alert to `status_message.<slack_channel>.#`. maybe a other naming convention.

### DIFF
--- a/bin/dragonfly
+++ b/bin/dragonfly
@@ -39,6 +39,7 @@ class DriplineParser(argparse.ArgumentParser):
                  amqp_broker=False,
                  config_file=False,
                  tmux_support=False,
+                 slack_support=False,
                  **kwargs):
         '''
         Keyword Args:
@@ -82,6 +83,14 @@ class DriplineParser(argparse.ArgumentParser):
                                           '--config',
                                           help='path (absolute or relative) to configuration file',
                                          )
+        if slack_support:
+            self.base_parser.add_argument('-s',
+                                          '--slack_channel',
+                                          help='define the Slack channel where to send the critical messages',
+                                          default=None,
+                                          nargs='?',
+                                          const='p8_alerts',
+                                         )
         if tmux_support:
             self.base_parser.add_argument('-t',
                                           '--tmux',
@@ -90,6 +99,7 @@ class DriplineParser(argparse.ArgumentParser):
                                           default=None, # value if option not given
                                           const=False, # value if option given with no argument
                                          )
+
         self.add_dragonfly_subcommands()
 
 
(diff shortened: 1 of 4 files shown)


wcpettus:
Why is defining slack a CLI option?  I would think that where we spam errors is fundamental enough this would be defined only in the config file.  Maybe I missed an earlier discussion on this though.

laroque:
Oh I understood this quite differently. I thought that the plan was to allow the slack to be configured so that "critical" message get sent to a different slack channel. I think it makes much more sense for a service to know/indicate the "severity" of a message (this is a status message, a warning, a critical error, etc.) and for the slack service to decide what it is supposed to do with that. It seems bad for any service other than the slack relay to need to know anything about slack, including what channels exist or are in use.

guiguem:
@wcpettus The CLI definition means "--slack_channel p8_alerts" will overwrite the one defined in the config file, such as we currently do for the broker.
I added it here so we have a testing tool: one can just use the same config file on a different slack channel (as we do for the broker) without editing the config file or/and committing it...

@laroque currently there is only one routine key ('status_message') where a message is sent to when the service does a logger.critical.
The routine key in the AMQPhandler should reworked in the format `logger.<logger_type>.<username>` (where `logger_type` would be critical, error, warning...) instead of `status_message.<channel_name>.<username>`. 
The scheme you propose does not allow a service to decide on which channel the message should be displayed.
For each mesh we would have a Slack interface that will only allow to send message to p8_<logger_type>, and cannot send something to let's say operator: in that sense it looses generality and flexibility.

To me, Slack interface should just send the messages where it is told so, instead of doing to selection itself, bypassing the actual need of the requesting service.
But I am open to discussion.

laroque:
I'm suggesting that the format should be `status_message.<severity>.<username>`, where "severity" may be "critical" or it may be "progress" or whatever. And then the slack service would know that "critical" messages go to #p8_critical or #p8_t2_critical and "progress" goes to #p8_notices (or a parallel one for t2 or maybe is ignored in slack) or however we want to configure it.

There could/should be a 1-to-1 mapping between the "severity" and the slack channel. We shouldn't actually use the slack channels in the service configurations because then 1) we have to reconfigure all of the services if we want to rename channels and 2) if we want to have some other class of services which also listens to these messages we're going to force the slack patterns onto it, even if they don't make sense there.

guiguem:
OK, I see. Then you would need to define this mapping in the Slack interface config file.
That seems fine.

guiguem:
What are the logger level we want every service to send messages for?
Maybe >=Warning?
As a reminder:
```
CRITICAL 50
ERROR	 40
WARNING	 30
INFO	 20
DEBUG	 10
NOTSET	 0
```

Issue #104, opened 2017-05-19
logger inspection


guiguem:
Need to look at every logger.XXX in dragonfly and dripline and make sure that each follows the following convention:
```
'warning'-> a user should know that something happened, they may or may not want to do something about it; 
'error' -> something broke, code is probably crashing and/or resetting itself, possible bug or similar problem, should probably create a github issue and/or contact a developer 
'critical' -> the code saw something that was explicitly designated as a concern, such as a value outside of its allowed range, a user should check on the system and probably should take corrective action
```
Need to add a documentation somewhere

Issue #47, opened 2016-07-27
set and check endpoint as a deeper level


guiguem:
A great enhancement of dragonfly would be to set and check the value of an endpoint, instead of just setting the endpoint, as a very deep level in the code. As it is currently done, we have to do this check in high-level classes such as the execution_script or the multiset. But if these self check were done in more basic function, we could only interact with the endpoint which will return a bool (and the value_raw and value_cal) according to the success of the operation.
Once endpoints support more generic self checking on set, the run_scripting’s set_and_check should migrate to use this.

wcpettus:
addendum - possibly allow a delay check option for slow responses
- power supplies take ~1 sec to reach the new set point (prompt check of the limit is still good)
- ion gauges take ~5 sec to come on (prompt check of the reply is still helpful)

buzinsky:
this is resolved too, no? or am I missing something (we do have a timeout argument for the last point)

wcpettus:
This issue is targeting a deeper-level change to how we set all endpoints in dragonfly - if any set automatically includes the relevant checks, then it would return a DriplineException of some flavor if the check failed.  Then we can strip out the checking from the MultiDo and scripting and do it right once.  It harkens back to the principle that if we don't trust the setting of our endpoints, we should fix it at the deepest level possible, and not slap a check bandaid higher-up.

PR #128, opened 2017-12-15
Feature/psyllid trigger

cclaessens:
This branch contains the methods for controlling psyllid in triggered mode.
The roach central frequency is limited to 1550MHz.
The psyllid interface and the roach_daq_interface can be started without setup calls. The system is prepared for taking data with "dragonfly cmd r2_channel_x_daq_interface.prepare_daq_system".


dragonfly/implementations/psyllid_provider.py
```diff
@@ -132,11 +135,10 @@ def reactivate(self, channel):
         if self.status_value_dict[channel] == 4:
             logger.info('Reactivating Psyllid instance of channel {}'.format(channel))
             self.provider.cmd(self.queue_dict[channel], 'reactivate-daq')
-            logger.info(result)
         else:
             logger.warning('Cannot reactivate Psyllid instance of channel {}'.format(channel))
             return False
-        time.sleep(1)
+        time.sleep(2)
```
laroque: Are we using this sleep to ensure there's time for the cmd to be processed? Should that be done by the time that line 137 returns? We don't need to delay the PR for it, but I'm pretty nervous about needing to hardcode sleeps like this.

dragonfly/implementations/psyllid_provider.py
```diff
@@ -233,51 +235,181 @@ def get_number_of_streams(self, channel):
...
+        return {'threshold': threshold, 'threshold_high' : threshold_high, 'n_triggers' : n_triggers}
+
+
+    # pre-trigger and skip-tolerance control
+    def set_time_window(self, pretrigger_time=2e-3, skip_tolerance=5e-3, channel='a'):
+        if self.mode_dict[channel] != 'triggered':
+            logger.warning('Psyllid not in triggered mode')
+            return False
+
+        self._set_pretrigger_time( pretrigger_time, channel)
```
laroque: you're not checking the return value of `_set_pretrigger_time`; We should either be sure to check the return values of functions that return a boolean to indicate their success condition, or change the functions not to return anything if we don't think it is useful to check them.

(as an aside, I think I generally prefer not returning booleans but instead raising exceptions if something goes wrong) Again, this doesn't need to prevent us from merging the PR

dragonfly/implementations/psyllid_provider.py
```diff
@@ -233,51 +235,181 @@ def get_number_of_streams(self, channel):
...
+    # internal psyllid time window and trigger parameter sets and gets
+    def _set_pretrigger_time(self, pretrigger_time, channel='a'):
+        n_pretrigger_packets = int(round(pretrigger_time/4.096e-5))
+        logger.info('Setting psyllid pretrigger to {} packets'.format(n_pretrigger_packets))
+        request = '.node-config.{}.eb.pretrigger'.format(str(self.channel_dict[channel]))
+        self.provider.set(self.queue_dict[channel]+request, n_pretrigger_packets)
         return True
 
 
+    def _get_pretrigger_time(self, channel='a'):
```
laroque: is there a reason that these get/set methods are all prefixed by `_`? That usually indicates that they are intended for internal use only. it isn't obvious to me why these couldn't be used externally, though maybe that is indeed the intent

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -264,29 +260,151 @@ def central_frequency(self):
...
+        if self.trigger_type != 'multi_trigger':
+            return False
+        result = self.provider.cmd(self.psyllid_interface, 'get_trigger_configuration', payload = self.payload_channel)
+        skip_tolerance = self.provider.cmd(self.psyllid_interface, 'get_time_window', payload = self.payload_channel)['skip_tolerance']
+
+        # In addition to threshold and n_triggers, returns skip_tolerance as it is n_triggers/skip_tolerance that defines this trigger
+        return {'threshold' : result['threshold'], 'n_triggers' : result['n_triggers'], 'skip_tolerance': skip_tolerance}
+
+
+    def set_multi_trigger(self, threshold, n_triggers):
```
laroque: if multi_trigger is an `@property` then probably `set_multi_trigger` should actually be:
```
@multi_trigger.setter
def multi_trigger(self, value):
    threshold, n_triggers = value
    ...
```

dragonfly/implementations/psyllid_provider.py
```diff
@@ -187,7 +189,7 @@ def set_central_frequency(self, channel, cf):
             return False
         except Exception as e:
```
wcpettus: I don't think this error handling is particularly consistent with the rest of the code for two reasons:
- we don't usually bother to catch generic exceptions
  - if we have a specific failure mode, catch the specific exception, logger.critical that this is a known case 
- this only raises a logger.error and then re-raises the same generic exception
  - hopefully whatever catches the exception finally will have all the information in the traceback
(sorry, this isn't new, and doesn't prevent the code from working)

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -183,7 +185,7 @@ def _start_data_taking(self, directory, filename):
         logger.info('Going to tell psyllid to start the run')
         payload = {'channel':self.channel_id, 'filename': os.path.join(directory, psyllid_filename), 'duration':duration}
         try:
-            self.provider.cmd(self.psyllid_interface, 'start_run', payload=payload)
+            self.provider.cmd(self.psyllid_interface, 'start_run', payload = payload)
         except core.exceptions.DriplineError:
             logger.error('Error from psyllid provider or psyllid')
```
wcpettus: if we are catching a specific error, it should probably be a logger.critical

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -264,29 +260,151 @@ def central_frequency(self):
...
+    @property
+    def single_threshold_trigger(self):
+        logger.info(self.trigger_type)
+        if self.trigger_type != 'single_threshold':
+            return False
+        threshold = self.provider.cmd(self.psyllid_interface, 'get_trigger_configuration', payload = self.payload_channel)['threshold']
+        return {'threshold' : threshold}
+
+
+    def set_single_threshold_trigger(self, threshold):
```
wcpettus: this is a simpler instance of Ben's below comment (requires no changes past the def line):
@single_threshold_trigger.setter
def single_threshold_trigger(self, threshold):

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -264,29 +260,151 @@ def central_frequency(self):
...
+
+    @property
+    def two_threshold_trigger(self):
+        if self.trigger_type != 'two_thresholds':
+            return False
+        result = self.provider.cmd(self.psyllid_interface, 'get_trigger_configuration', payload = self.payload_channel)
+        return {'low_threshold' : result['threshold'], 'high_threshold' : result['threshold_high']}
+
+
+    def set_two_threshold_trigger(self, high_threshold, low_threshold):
```
wcpettus: as above, cleaner as an @property .setter

dragonfly/implementations/psyllid_provider.py
cclaessens: We have to wait with checking psyllid's status until it has finished changing that. Otherwise the returned status will be activating or deactivating instead of activated.
In psyllid/source/control/daq_control, reactivate is simply deactivate followed by activate. Psyllid itself waits for a hardcoded second between those too calls because it would otherwise run into the same problem:

    void daq_control::reactivate()
    {
        deactivate();
        std::this_thread::sleep_for( std::chrono::seconds(1) );
        activate();
        return;
    }

We could not check the status after telling psyllid to deactivate, activate or reactivate. The status switching in psyllid seems pretty reliable by now.
Would you prefer that?

dragonfly/implementations/psyllid_provider.py
cclaessens: one of the returns was actually checked in set_trigger_configuration. The idea was that psyllid_provider returns True or False for everything (unless something unexpected happens) and the roach_daq_interface gets to decide whether the return should result in an error.
I currently don't see a reason why that couldn't be changed, so I will remove the boolean returns and raise errors in the psyllid_provider directly.

dragonfly/implementations/psyllid_provider.py
cclaessens: They could be used externally. Initially they were not prefixed by _ and I did use them to set and get trigger settings. But as the number of configurable parameters grew I became lazy and added set_trigger_configuration (sets 4 trigger parameters) and set_time_window (sets pretrigger and skip tolerance). I only use these now and the roach_daq_interface also only uses these. Therefore I added the prefixes to make them internal.
In execution scripts it probably doesn't matter whether you have 6 sets or 2. You set 6 parameters either way. Would you prefer to have the 6 set methods for external use?

dragonfly/implementations/roach_daq_run_interface.py
cclaessens: I totally agree but don't know how to make an attribute setter that takes more than 1 argument.
Can you tell me how to do that?

dragonfly/implementations/psyllid_provider.py
cclaessens: Is the difference between logger.critical and logger.error that logger.critical are known errors?

cclaessens: for now I added a variable self.mode_testing that distinguishes between testing the acquisition mode and general frequency setting.

dragonfly/implementations/psyllid_provider.py
cclaessens: Also:
The trigger and frequency settings are active node configurations. 
When psyllid is reactivated they are lost and have to be reset. You can only set them when psyllid is activated. With the hardcoded sleep we make sure that new active node settings cannot be applied while psyllid is still reactivating.

dragonfly/implementations/roach_daq_run_interface.py
laroque: Oh, you'd have to either pack the pair of values into a tuple that could be unpacked (that is, if `value == (threshold_val, n_triggers_val)`.... I think that that should be doable from code, but the CLI for dragonfly probably doesn't support it.

The way drip line would prefer it done would be to split them into multiple properties. One for the threshold and another for the number of triggers, each with some default value. That probably requires some refactoring so I'll let it go unless you think that sounds easy.

wcpettus: I've had pretty spotty luck with CLI parsing of lists and things.  If we need this we could do some testing of dripline_agent to get the desired behavior.

dragonfly/implementations/psyllid_provider.py
wcpettus: A broad code review was proposed months ago to straighten out our use of different logger settings, but we haven't done it:
https://github.com/project8/dragonfly/issues/104

The goal for stable running is that we get no logger.critical (or logger.error) messages; if we get such a problem we need to stop and intervene, assess what went wrong.  We use logger.critical to notify of problems we've seen before but hope not to see during normal running - the coldhead is warming, or an ethernet service goes unresponsive.  Usually you don't need to catch generic exceptions because they are caught by other parts of the code.

Logger.warning can be ok to receive during stable running, but should be notified.  We use them when we mute pingers, when a SQL snapshot returns empty, or when you attempt to get/set an endpoint that doesn't have that functionality.

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -183,9 +185,9 @@ def _start_data_taking(self, directory, filename):
         logger.info('Going to tell psyllid to start the run')
         payload = {'channel':self.channel_id, 'filename': os.path.join(directory, psyllid_filename), 'duration':duration}
         try:
-            self.provider.cmd(self.psyllid_interface, 'start_run', payload=payload)
+            self.provider.cmd(self.psyllid_interface, 'start_run', payload = payload)
         except core.exceptions.DriplineError:
-            logger.error('Error from psyllid provider or psyllid')
+            logger.critical('Error from psyllid provider or psyllid')
```
guiguem: Can we print the error?

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -195,7 +197,7 @@ def _start_data_taking(self, directory, filename):
             logger.error('Local error')
```
guiguem: More details?

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -210,49 +212,43 @@ def _stop_data_taking(self):
...
             if self.status_value==None:
                 raise core.exceptions.DriplineGenericDAQError('Psyllid must have crashed during run')
 
 
-    def adopt_new_psyllid_instance(self):
-        self.provider.cmd(self.psyllid_interface, 'get_acquisition_mode', payload = self.payload_channel)
-        self._finish_configure()
-
-
     def stop_psyllid(self):
```
guiguem: return something like 'psyllid stopped'

dragonfly/implementations/psyllid_provider.py
```diff
@@ -233,51 +238,167 @@ def get_number_of_streams(self, channel):
...
+        logger.info('Switch tf_roach_receiver back to time and freq')
+        request = 'run-daq-cmd.{}.tfrr.time-and-freq'.format(str(self.channel_dict[channel]))
         result = self.provider.cmd(self.queue_dict[channel],request)
 
-        logger.info('apply-trigger')
-        request = 'run-daq-cmd.ch'+str(self.channel_dict[channel])+'.fmt.apply-trigger'
+        logger.info('Switch frequency mask trigger to apply-trigger')
+        request = 'run-daq-cmd.{}.fmt.apply-trigger'.format(str(self.channel_dict[channel]))
         result = self.provider.cmd(self.queue_dict[channel],request)
-        return True
```
guiguem: Can we return something like "Awesome, the mask has been generated"?

dragonfly/implementations/psyllid_provider.py
```diff
@@ -233,51 +238,167 @@ def get_number_of_streams(self, channel):
...
-        request = 'start-run'
-        result = self.provider.cmd(self.queue_dict[channel], request)
+        logger.info('Start short run to record mask')
+        self.start_run(channel ,1000, '/home/project8/fmt_update.egg')
         time.sleep(1)
 
-        logger.info('write-mask')
-        request = 'run-daq-cmd.ch'+str(self.channel_dict[channel])+'.fmt.write-mask'
+        logger.info('Write mask to file')
+        request = 'run-daq-cmd.{}.fmt.write-mask'.format(str(self.channel_dict[channel]))
```
guiguem: Or maybe the trigger mask writting is more important here?

dragonfly/implementations/psyllid_provider.py
```diff
@@ -37,9 +37,10 @@ def __init__(self,
         self.channel_dict = {'a': 'ch0', 'b': 'ch1', 'c': 'ch2'}
         self.freq_dict = {'a': None, 'b': None, 'c': None}
         self.mode_dict = {'a':None, 'b':None, 'c':None}
+        self.mode_testing = False
```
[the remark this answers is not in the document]
cclaessens: I am not sue I understand what you mean.
Channel a, b and c are hardcoded in r2daq as well. We could always just use a single channel and call that a. If not we have to change things in the roach2_interface and r2daq. But I think that is independent of the dictionaries in psyllid_provider. ch0, ch1 and ch2 are hardcoded again in psyllid and just naming conventions. The ports that correspond to channel a, b and c on the roach and ch0, ch1 and ch2 are already defined in the config files.
So the flexibility of using 1 to 3 channels and just needing to change config files is already there.

dragonfly/implementations/psyllid_provider.py
```diff
@@ -37,9 +37,10 @@ def __init__(self,
...
 
 
-    def _finish_configure(self):
+    def check_all_psyllid_instances(self):
```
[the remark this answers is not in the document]
cclaessens: We can remove this method if you like. It used to be the setup call method but we decided on the call to not do setup calls anymore. The operator could use it if he wanted to check how many psyllid instances are running and which instances are in triggering or in streaming mode.
But he can also ask the r2_channel_x_interfaces whether the corresponding psyllid instance is in triggering or streaming mode. So it is actually redundant.
What do you say? Should we just remove it?

dragonfly/implementations/psyllid_provider.py
```diff
@@ -53,14 +54,13 @@ def _finish_configure(self):
 
                 except core.exceptions.DriplineGenericDAQError:
                     logger.warning('No matching Psyllid instance for channel {} present'.format(channel))
```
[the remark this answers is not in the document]
cclaessens: The channels are blocked if the mode of frequency dictionary for this channel says None. So this is already taken care of. Again, we don't need this method to do that. When psyllid_provider starts, it says None in the dictionaries everywhere and when the operator then calls: r2_channel_x_daq_interface.prepare_daq_system the dictionaries are filled for this channel with the true mode and frequency and data taking can start.

dragonfly/implementations/psyllid_provider.py
```diff
@@ -182,12 +185,14 @@ def set_central_frequency(self, channel, cf):
...
         except core.exceptions.DriplineError:
-            logger.error('Could not set central frequency of Psyllid instance for channel {}'.format(channel))
-            self.freq_dict[channel]=None
-            return False
-        except Exception as e:
-            logger.error('Unexpected error in set_central_frequency')
-            raise 
+            if self.mode_testing == True:
+                self.freq_dict[channel]=None
+                logger.info('Could not set central frequency of Psyllid instance for channel {}'.format(channel))
```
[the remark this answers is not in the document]
cclaessens: normally yes. 
but if self.mode_testing itsTrue it is not. Because then *get_acquisition_mode* in psyllid_interface is testing which settings can be applied to psyllid. Otherwise the failure to set the frequency results in a logger.critical message.

dragonfly/implementations/psyllid_provider.py
```diff
@@ -182,12 +185,14 @@ def set_central_frequency(self, channel, cf):
...
+                return False
+            else:
+                logger.critical('Could not set central frequency of Psyllid instance for channel {}'.format(channel))
+                self.freq_dict[channel]=None
+                return False
 
 
     def start_run(self, channel, duration, filename):
```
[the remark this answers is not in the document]
cclaessens: yes, but why? If it doesn't work there will be an error from psyllid.

dragonfly/implementations/psyllid_provider.py
```diff
@@ -233,51 +238,167 @@ def get_number_of_streams(self, channel):
...
-            logger.warning('Psyllid not in streaming mode')
-            return False
-        request = '.node-config.ch'+str(self.channel_dict[channel])+'.fmt.threshold-power-snr'
-        result = self.provider.set(self.queue_dict[channel]+request, snr)
-        self.deactivate(channel)
-        time.sleep(1)
-        self.activate(channel)
-        return True
+    # trigger control
+    def set_trigger_configuration(self, channel='a', threshold=16, threshold_high=0, n_triggers=1,):
```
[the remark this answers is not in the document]
cclaessens: They are not used by someone. They are used by the roach_daq_run_interface. If you *set_trigger_configuration* there it will first call *psyllid_interface.set_trigger_configuration* and then *psyllid_interface.get_trigger_configuration* and return the result. The operator will see what settings were applied.

dragonfly/implementations/psyllid_provider.py
```diff
@@ -233,51 +238,167 @@ def get_number_of_streams(self, channel):
...
+    def make_trigger_mask(self, channel='a', filename='~/fmt_mask.json'):
+        if self.mode_dict[channel] != 'triggered':
+            logger.error('Psyllid instance is not in triggered mode')
+            raise core.exceptions.DriplineGenericDAQError("Psyllid instance is not in triggered mode")
 
-        logger.info('freq-only')
-        request = 'run-daq-cmd.ch'+str(self.channel_dict[channel])+'.tfrr.freq-only'
+        logger.info('Switch tf_roach_receiver to freq-only')
+        request = 'run-daq-cmd.{}.tfrr.freq-only'.format(str(self.channel_dict[channel]))
         result = self.provider.cmd(self.queue_dict[channel],request)
```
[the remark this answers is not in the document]
cclaessens: if there is an error in the procedure psyllid will tell you

dragonfly/implementations/psyllid_provider.py
```diff
@@ -233,51 +238,167 @@ def get_number_of_streams(self, channel):
...
+        logger.info('Switch frequency_mask_trigger to update-mask')
+        request = 'run-daq-cmd.{}.fmt.update-mask'.format(str(self.channel_dict[channel]))
         result = self.provider.cmd(self.queue_dict[channel],request)
         time.sleep(1)
 
-        logger.info('run')
-        request = 'start-run'
-        result = self.provider.cmd(self.queue_dict[channel], request)
+        logger.info('Start short run to record mask')
+        self.start_run(channel ,1000, '/home/project8/fmt_update.egg')
```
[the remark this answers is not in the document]
cclaessens: this is not the trigger mask. the trigger mask is stored here:

        timestr = time.strftime("%Y%m%d_%H%M%S")
        filename = '{}_frequency_mask_channel_{}_cf_{}.json'.format(timestr, self.channel_id, self.freq_dict[self.channel_id])
        path = os.path.join(self.mask_target_path, filename)

*mask_target_path* has to be set in the config file of *r2_channel_x_daq_interface*.

To record a mask psyllid has to start a run in frequency-only mode. The mask that is recorded is written to a json file. But currently psyllid cannot start a run without creating an egg file. This file is empty and has no use. Therefore I thought I give it a hardcoded destination where it will always be overwritten. Where do you want that to be?

dragonfly/implementations/psyllid_provider.py
cclaessens: ok :)

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -65,6 +67,14 @@ def _finish_configure(self):
         self.central_frequency = freqs[self.channel_id]
 
 
+    def prepare_daq_system(self):
+        acquisition_mode = self.provider.cmd(self.psyllid_interface, 'get_acquisition_mode', payload = self.payload_channel)
+        if acquisition_mode['values'][0] == None:
```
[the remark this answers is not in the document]
cclaessens: yes, unless there is no psyllid instance running. In that case the mode is None and an error is raised.

dragonfly/implementations/roach_daq_run_interface.py
cclaessens: ok

dragonfly/implementations/roach_daq_run_interface.py
cclaessens: the error is re-raised after the channel was unblocked. this would be an unexpected error and I don't know more details about it. I dont think this exception has ever been caught so far

dragonfly/implementations/roach_daq_run_interface.py
cclaessens: ok

dragonfly/implementations/psyllid_provider.py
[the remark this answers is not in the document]
laroque: If it works then I don't think we can just leave it. The daq interface is here to mostly prevent "a user" from needing to go talk to the r2_* objects directly for "normal" use. If at some point the method is broken then we can consider removing it rather than maintaining it.

The docstring comment is a valid one, but is pervasive. It would be useful to add at least a short docstring for each function. I think we can merge this PR, add an open issue to add those, and let Christine add/merge those directly. That way the code changes can be deployed whenever we're ready to re-deploy.

dragonfly/implementations/psyllid_provider.py
laroque: I guess the point for the sleep is well made if that's the psyllid code. It isn't obvious to me that things are self-consistent, but that's maybe not a question for here.

dragonfly/implementations/psyllid_provider.py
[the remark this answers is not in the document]
laroque: I disagree with the statement that there should be a return. A set function should not return anything. It should assign the value. If there is an error it should raise an exception, no exception implies success.

dragonfly/implementations/psyllid_provider.py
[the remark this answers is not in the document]
laroque: I agree with Christine. We should only add a return value if there is some useful piece of data to return (do we need to return the RID for example).

dragonfly/implementations/psyllid_provider.py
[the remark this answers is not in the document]
laroque: If the result isn't being used, you can replace `result = self.provider.cmd(<stuff>)` with `self.provider.cmd(<stuff>)` and let the result be garbage collected before placing it into the local scope.

When you say "psyllid will tell you" do you mean it will send a message for the user? If psyllid sends a reply with a non-zero error code to one of the first cmd requests, doesn't that mean that there is an error and making the mask has failed? This should raise an exception here and return an error code to whatever requested it of this service.

dragonfly/implementations/psyllid_provider.py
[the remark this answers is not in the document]
laroque: @nsoblath how hard would it be for psyllid to be changed to not create this egg file? If we're going to create a dummy file we should put it into a place where we know it will be cleaned up (and ideally do that ourselves). The default should probably in `/tmp/` and the path for it configurable, that way we can send it somewhere for dung beetle or hornet or something to clean up if we decide we need that.

dragonfly/implementations/psyllid_provider.py
laroque: As above, I disagree with having non-informative return values. We should be consistent with the rest of the code. Unless there is a specific reason we want to try to do something and keep going if it fails, then we should raise exceptions. If such a condition exists then the function name should reflect that. Something like `make_trigger_mask_and_check()` or `check_make_trigger_mask()`.

dragonfly/implementations/roach_daq_run_interface.py
laroque: you can `print(e.message)` to print the message content of the exception or print `('exception while <whatever you were doing>, trying to cleanup')`.

dragonfly/implementations/roach_daq_run_interface.py
laroque: as above, don't return anything; our convention is to use exceptions and reply return codes to indicate success failure.

dragonfly/implementations/psyllid_provider.py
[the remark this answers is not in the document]
laroque: Actually that's not even true. The implementation of `provider.cmd` should already be checking the return code and raising the exception. you should just not catch the `result` into a local unless the cmd produces some useful content in the payload which you need to use (which doesn't appear to be the case here anywhere).

dragonfly/implementations/psyllid_provider.py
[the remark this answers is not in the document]
wcpettus: You have defined a naming convention and mapping that could be be defined in the config file as:
```yaml
channel dict:
    a: ch0
    b: ch1
    c: ch2
```
And then here:
```python
         self.channel_dict = channel_dict
         self.freq_dict = {x: None for x in channel_dict}
         self.mode_dict = {x: None for x in channel_dict}
```
This makes the code more generic because you've removed hard-coded parameters from dragonfly.  If psyllid is ever changed because we decide its confusing to use a,b,c here and 0,1,2 there, the change is swept up in the config file.  If we get a new digitizer board that has 8 channels, the same code adapts with only a different config file (also you don't have to replace every instance of your a-c dicts, which could get tedious.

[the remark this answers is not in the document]
guiguem: Thanks, this is what I meant!

dragonfly/implementations/psyllid_provider.py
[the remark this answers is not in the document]
wcpettus: If we ever try to do set_and_check (https://github.com/project8/dragonfly/issues/47), everything will be terrible.

_This is way out of scope for this PR._

dragonfly/implementations/psyllid_provider.py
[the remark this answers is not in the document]
guiguem: is this testing mode something we want in the release? I understand this is useful for testing, but do we need it?

dragonfly/implementations/psyllid_provider.py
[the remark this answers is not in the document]
guiguem: As defined here, this will be used via "dragonfly cmd" and I thought we want cmd to return stuff.
If this was something used via 'dragonfly set' I would agree, but this is not (or at least not as currently implemented).

dragonfly/implementations/roach_daq_run_interface.py
wcpettus: @laroque: is there a reason to print the message instead of logger.info it?

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -183,19 +185,19 @@ def _start_data_taking(self, directory, filename):
...
         try:
-            self.provider.cmd(self.psyllid_interface, 'start_run', payload=payload)
-        except core.exceptions.DriplineError:
-            logger.error('Error from psyllid provider or psyllid')
+            self.provider.cmd(self.psyllid_interface, 'start_run', payload = payload)
+        except core.exceptions.DriplineError as e:
+            logger.critical('Error from psyllid provider or psyllid. Starting psyllid run failed')
             payload = {'channel': self.channel_id}
             try:
                 self.provider.cmd(self.daq_target, 'unblock_channel', payload = payload)
```
wcpettus: Are there any cases when a channel will already be blocked when entering this `_start_data_taking` method?  Because you don't check if the channel is blocked before blocking, the `daq_target.block_channel` could do nothing (makes it True, when it was already True), and then unblocking may be problematic?  I don't know what checks have already been made before calling this, maybe this can't happen.

Specific use case: What if the `psyllid_interface.start_run` fails because a run is already going, then you wouldn't want to unblock the channel because there is still a legit run going. (I don't know if that can happen though).

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -210,49 +212,44 @@ def _stop_data_taking(self):
...
         except Exception as e:
             logger.error('Local error at end of run')
             logger.info('unblock channel')
             payload = {'channel': self.channel_id}
             try:
-                self.provider.cmd(self.daq_target, 'unblock_channel', payload=payload)
+                self.provider.cmd(self.daq_target, 'unblock_channel', payload = payload)
             finally:
                 raise e
         else:
```
wcpettus: this else doesn't close an if statement - should it be up at line 214?

dragonfly/implementations/psyllid_provider.py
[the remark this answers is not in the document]
nsoblath: I added a feature to psyllid that lets one turn off the use of egg files by sending a dl `set` command to `psyllid.use-monarch`.  `false` will turn it off, and `true` will turn it back on.

If it's set to `false` and a writer attempts to write to a file, psyllid will crash.

dragonfly/implementations/psyllid_provider.py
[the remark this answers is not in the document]
cclaessens: we need it when psyllid_provider is testing whether psyllid is in streaming or triggering mode. it is generally set to false and then failing to set the frequency will result in an error.

dragonfly/implementations/roach_daq_run_interface.py
cclaessens: start_run calls _do_checks before calling _start_data_taking.
_do_checks already checks whether the channel is blocked (and raises an error if it is) so _start_data_taking doesn't need to check again.
_do_checks also checks whether there is a run ongoing.

PR #137, opened 2018-02-20
Release/v1.11.0

guiguem:
A release that includes a bunch of things:
- a triggering ROACH
- RSA DAQ interface separated from the base class
- some documentation (in progress)
- the mdreceiver ported in python
- other stuff?


guiguem:
A lot of the remaining code changes would require to change internal methods into staticmethods because we don't use the self.
Is it worth the effort to try and fixe these?

PR #157, opened 2018-07-12
new spime to read multiple adc's and calculate based on them

wcpettus:
This is the quick-and-dirty implementation of a new Spime to do some calculation based on multiple ADC readings.  The logic:
- value_raw will return a calculation based on multiple readings (kind-of-raw, I want a ratio of voltages)
- value_cal will return this calibrated into what you care about (calibrated temperature from a thermocouple)

Is there a cleaner implementation?
- I also considered making a new `ADS1115Read` class which both the `ADS1115Spime` could straightforwardly inherit from and `ADS1115CalcSpime` could initialize N of, then the reads would call to there.  I don't think there's a parallel to this in the code and that could start looking messy fast.
- The CalcSpime could initialize two Spimes with logging off to handle the two reads, but that's a lot of overhead, and all the provider/service linkages in Spime/Scheduler would be broken.

### DIFF
--- a/dragonfly/implementations/spime_endpoints.py
+++ b/dragonfly/implementations/spime_endpoints.py
@@ -11,6 +11,7 @@ Spime catalog (in order of ease-of-use):
 -* SimpleSCPIGetSpime/SimpleSCPISetSpime: limited instance of above with disabled Get/Set
 - FormatSpime: utility spime with expanded functionality
 - ADS1115Spime: spime to read ADS1115 16-bit ADC
+- ADS1115CalcSpime: spime to perform complex logic on multiple ADS1115 readings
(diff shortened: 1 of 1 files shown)


laroque:
Was this tested and working? I think that this level of logic is a place where the dripline-python conventions could take a page from dripline-cpp. Both for describing services, and run scripting, I think it is usually nice if one achieve most tasks by writing just a config file. But I may have pushed that ideal too hard, there are some times where it is easier to just write actual code to implement the thing you want and then be done with it. In the future I think that that will probably live with the config file rather than in the repo, but for now we should keep going with it as is.

wcpettus:
Yes - this is tested to work for the coldhead system, should have mentioned that.

PR #158, opened 2018-08-01
Feature/default threshold type

cclaessens:
- removed exception raising in psyllid_provider if acquisition_mode is None.
- replaced a trigger_type check with acquisition_mode check in roach_daq_run_interface (which by itself would resolve the error spamming issue because that check is not made on a state but actually asks psyllid_provider which then knows the mode)
- added threshold type to default trigger settings.

Do we also want to update the acquisition_mode check, so that dragonfly gets psyllid's node config?

I haven't tested this branch yet.

### DIFF
--- a/dragonfly/implementations/roach_daq_run_interface.py
+++ b/dragonfly/implementations/roach_daq_run_interface.py
@@ -488,6 +488,8 @@ class ROACH1ChAcquisitionInterface(DAQProvider):
         '''
         Returns pretrigger time and skip tolerance
         '''
+        if self.acquisition_mode is not 'triggering':
+            return False
         result = self.provider.cmd(self.psyllid_interface, 'get_time_window', payload = self.payload_channel)
         return result
 
--- a/dragonfly/implementations/roach_daq_run_interface.py
+++ b/dragonfly/implementations/roach_daq_run_interface.py
@@ -527,6 +527,7 @@ class ROACH1ChAcquisitionInterface(DAQProvider):
         if self.default_trigger_dict == None:
             raise core.exceptions.DriplineGenericDAQError('No default trigger settings present')
         else:
+            self.threshold = self.default_trigger['threshold-type']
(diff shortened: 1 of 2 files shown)


dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -463,7 +463,7 @@ def trigger_settings(self):
         '''
         Returns all trigger settings
         '''
-        if self.trigger_type == None:
+        if self.acquisition_mode == None:
```
wcpettus: shouldn't this check that acquisition_mode is 'triggering'?  if in streaming mode trigger settings won't exist because the fmt node doesn't exist?

wcpettus: also do this same check at line 491?

wcpettus:
For your last comment are you suggesting modifying https://github.com/project8/dragonfly/blob/feature/default-threshold-type/dragonfly/implementations/psyllid_provider.py#L102 method to more directly get the mode from psyllid's node config?

dragonfly/implementations/roach_daq_run_interface.py
```diff
@@ -527,6 +527,7 @@ def set_default_trigger(self):
         if self.default_trigger_dict == None:
             raise core.exceptions.DriplineGenericDAQError('No default trigger settings present')
         else:
+            self.threshold_type = self.default_trigger['threshold-type']
```
wcpettus: should be `self.default_trigger_dict`

cclaessens:
currently get trigger_settings returns False when the acquisition mode is not triggering but get time_window_settings results in an error when the channel is streaming and is therefore not consistent. which is the behavior we want?

cclaessens:
@wcpettus just saw your comments and added the mode check to line 491.

cclaessens:
and yes, in my first comment on this pr, I was referring to line 102 in psyllid_provider.

wcpettus:
Yes definitely to changing psyllid_provider.  The old way was the best we could do, but that new functionality is a much saner way of testing the acquisition mode.

PR #162, opened 2018-10-23
Feature/more generic table

laroque:
Making the SQL table interfaces a bit more adaptable. Testing admx control teststand with their database structure.


wcpettus:
We should also use six more broadly for string checks throughout the code.

laroque:
opened an issue reflecting the above comment. I don't think either of us are going to get to that now, at least now we may find it when skimming the issues to find good projects for new contributors.

PR #164, opened 2018-11-02
Feature/dungbeetle

laroque:
I'm making a PR mostly on behalf of @yadiw. We have some updates to discuss before external review or merge.

### DIFF
--- a/dragonfly/implementations/dungbeetle.py
+++ b/dragonfly/implementations/dungbeetle.py
@@ -6,8 +6,8 @@ __all__ = []
 __all__.append('DungBeetle')
 @fancy_doc
 class DungBeetle(Endpoint,Scheduler):
-
-    def __init__(self, root_dirs = [], max_age = "2h", ignore_dirs = [], **kwargs):
+    def __init__(self, root_dirs = [], max_age = "2h", 
+					ignore_dirs = [], **kwargs):
         Endpoint.__init__(self, **kwargs)
         Scheduler.__init__(self, **kwargs)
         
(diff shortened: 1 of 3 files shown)


laroque:
I'll make some comments on particular lines, but first a general comment:
The source currently mixes spaces and tabs; since python uses white space for syntax, it is critical that we be very consistent. Please convert all of the tab characters to spaces. I also recommend changing the settings of your editor to use spaces instead of tabs (while still providing the automatic indentation feature that makes your life easier). Depending on the editor I may or may not know how to help with that.

dragonfly/implementations/dungbeetle.py
```diff
@@ -0,0 +1,53 @@
...
+    # clean up empty directories under a specific directory without deleting itself
+    def clean_dir(self):
+        min_ctime = datetime.datetime.now() - datetime.timedelta(seconds = self.get_seconds(self.max_age))
+	for root_dir in self.root_dirs:
+            if os.path.isdir(root_dir):
+                for item in os.listdir(root_dir):
+                    sub_path = os.path.join(root_dir, item)
+                    self.del_dir(sub_path, min_ctime)
+            else:
+                raise Exception(" path [{}] does not exist.".format(root_dir))
```
laroque:
```suggestion
                raise ValueError(" path [{}] does not exist or is not a directory".format(root_dir))    
```

We will reach this else block if the path exists but is not a directory, let's update the exception text to be a bit more clear. I don't think we need to fully diagnose the problem for the user beyond this. It is also nice to raise a more specific exception, if there's on available (
https://docs.python.org/2/library/exceptions.html#exceptions.ValueError)

laroque:
I've created https://github.com/project8/dragonfly/pull/165 with further changes. Please review and merge if you're happy with that (otherwise we can discuss, that should update this PR when those changes are in).

PR #170, opened 2018-11-15
fix warning level problem

yadiw:
- now directories with sub-directories don't send a warning any more.
 - when a certain directory triggers the same warning multiple times, it will now send it normally in the first 5 cycles, and then one warning per 5 cycles.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,6 +12,7 @@ Luis Saldana [5]
 Matthew Sternberg [3]
 Luiz de Vivieros [4]
 Megan Wachtendonk [3]
+yadiw [3]
 
 
 Institutions
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,7 +12,7 @@ Luis Saldana [5]
 Matthew Sternberg [3]
 Luiz de Vivieros [4]
 Megan Wachtendonk [3]
-yadiw [3]
+Yadi Wang [3]
 
 
 Institutions
--- a/dragonfly/implementations/dungbeetle.py
+++ b/dragonfly/implementations/dungbeetle.py
@@ -28,33 +28,26 @@ class DungBeetle(Endpoint,Scheduler):
         self.root_dirs = root_dirs
         self.max_age = datetime.timedelta(**max_age)
         self.ignore_dirs = ignore_dirs
-        self.processed_dirs = {}
 
     # recursively delete empty directories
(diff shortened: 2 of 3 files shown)


laroque:
Hey Yadi,

This looks better, but I think we may have an even nicer solution. We can pull the logger.warning out of the loop and instead collect the list of directories that meet this condition; then either at the end of `clean_dir`, or in `scheduled_action` after the call to `clean_dir`, if the list is not empty we can send a single warning which says something like `"paths {} not removed because not empty".format(not_removed_paths)`. This way we only generate a single warning, regardless of the number of directories that were unable to be removed.

Hopefully @wcpettus can comment on if we want this to happen on each cycle or if we want some acknowledgement/backoff system. If so we will probably want to do something more sophisticated. A couple of options/questions are:
1. could the backoff happen in the slack relay? (I haven't looked at that code in a while so maybe not)
2. follow the pinger's model and silence repeated alarms until some configured future time (use the same max behavior as the pinger... do we need to find a way to abstract this? that was the motivation for 1, but it seems challenging)
3. keep something similar to the existing solution, which alarms some configured number of times in a row, then reduces the frequency if they continue. I don't think I like a self-silence without some user first acknowledging the situation.
4. If we've gotten behind, do we want each directory which is newly unable to be removed to create an alarm, even if we've done something to silence them? I think that if there are generally problems with removal, each case is probably not special, but maybe I'm missing something.

laroque:
Discussing with Yadi in person, our path forward is:
- collapse to 1 warning per cycle (this is either done or in progress)
- on each cycle if there is a *new* directory trying to create a warning, send a logger.warning
- if the set of directories with problems is not empty, but has nothing new, send a logger.warning every N cycles (where N is configured in the config file)

This lets us avoid extra backoff abstraction and avoids hard-coding some choice. Looking at the sensor monitor, we have an example of sending an alarm every 1 hour, so whatever that is in number of cycles could be a good starting point. It is also concrete expectations for Yadi to implement.

yadiw:
I've made changes to the dungbeetle program and did a simple testing on my computer, now it should work as what Ben describes.
I also added the new parameter to the example file- please feel free to correct me if there's anything wrong : )

wcpettus:
This looks complete and ready to merge.  I'm going to let it sit for a day until I have time to deploy the hotfix after tagging.

PR #176, opened 2019-03-01
Feature/at operator integration

yadiw:
Basically the same as the Go Operator.
- call the operator listed on Google calendar when someone do "@operator" in any project 8 slack channel (an all day shift is interpreted as 9am-9am).
- (exactly!) same helper commands available.
- regularly send messages when the bot gets started / the operator changes / it's being terminated, if a valid "monitor channel" is specified.
- regularly update information such as Google calendar events / slack users.

These are all the functions I can think of now. Feel free to make changes if you want it to send some more interesting messages :)


buzinsky:
This correctly got the operator for both over the weekend (none), and going into the new week (Tim).

laroque:
@buzinsky will you have a chance to review the code itself soon?

dragonfly/implementations/at_operator.py
```diff
@@ -0,0 +1,460 @@
...
+
+        else:
+            return None, None
+
+    def at_operator(self, channel):
+        '''
+        Send a Slack message to @ the operator (@ temporary operator(s) if exists, otherwise @ the one listed on google calendar.)
+        channel: id of the channel where the message will be sent
+        '''
+        self.send_message(channel, "Get it!")
```
wcpettus: we should probably just remove this message in production?

dragonfly/implementations/at_operator.py
```diff
@@ -0,0 +1,460 @@
...
+                        num_args = len(funcsigs.signature(func).parameters)
+                        args = [channel, user_id, operator_name]
+                        func(*args[:num_args])
+        return None
+
+    def run(self):
+        '''
+        The main loop.
+        '''
+        logging.getLogger('googleapiclient.discovery_cache').setLevel(logging.ERROR)
```
wcpettus: Is this what causes the logging thread to duplicate everything?

dragonfly/implementations/at_operator.py
```diff
@@ -0,0 +1,460 @@
...
+        logger.info(" Trying to construct an dictionary mapping channel names to their ids.")
+        request = self.slack_client.api_call("conversations.list")
+        if request['ok']:
+            channel_name_to_id_dictionary = {}
+            channel_id_to_name_dictionary = {}
+            for c in request['channels']:
+                channel_name_to_id_dictionary[c['name']] = c['id']
+                channel_id_to_name_dictionary[c['id']] = c['name']
+            logger.info(' Constructed dictionaries containing information of ' + str(len(request['channels'])) + ' channels.')
+            return channel_name_to_id_dictionary, channel_id_to_name_dictionary
```
wcpettus: when I tried accessing the dictionaries they were blank, I'm unsure why given that the log said `len(request['channels'])` == 73

dragonfly/implementations/at_operator.py
```diff
@@ -0,0 +1,460 @@
...
+        '''
+        logger.info(" Trying to construct an dictionary mapping channel names to their ids.")
+        request = self.slack_client.api_call("conversations.list")
+        if request['ok']:
+            channel_name_to_id_dictionary = {}
+            channel_id_to_name_dictionary = {}
+            for c in request['channels']:
+                channel_name_to_id_dictionary[c['name']] = c['id']
+                channel_id_to_name_dictionary[c['id']] = c['name']
+            logger.info(' Constructed dictionaries containing information of ' + str(len(request['channels'])) + ' channels.')
```
wcpettus: cleaner using .format() notation?

dragonfly/implementations/at_operator.py
```diff
@@ -0,0 +1,460 @@
...
+                logger.critical(" Cannot get the list of users.")
+                os._exit(1)
+
+            self.channel_name_to_id_dictionary, self.channel_id_to_name_dictionary = self.construct_channel_dictionaries()
+            if not self.channel_name_to_id_dictionary:
+                logger.error(' Cannot get the list of channels, therefore cannot find the monitor channel. Related functionalities will be disabled.')
+            elif self.monitor_channel_name not in self.channel_name_to_id_dictionary:
+                logger.error(' The name of monitor channel is not found. Related functionalities will be disabled.')
+            else:
+                self.monitor_channel_id = self.channel_name_to_id_dictionary[self.monitor_channel_name]
```
wcpettus: the `monitor_channel_id` variable is a blank string, the logger from the next line indicated the coded did reach this else

wcpettus:
Flaw in execution:
- the dripline "service" can crash but the program doesn't crash, this breaks all dripline communication while the slack + GCal bits keep working; is this because of the while/try loop construction? running as daemon we don't need the keyboard interrupt anyway

This looks like magic confusion:
- when I test the value of `self.monitor_channel_id`, it's a blank string; I don't understand how it correctly routes its operator shift change messages given this is the case

yadiw:
> Thanks for the changes. This looks very good. I do have two comments:
> 
> I think the run() function should be split up a bit more, it is a bid unwieldy to read. Functions should do 1 thing well: could split between the pre-run tests/ logging and actual while loop
> "rocognize" in remove tmpoperator function
Fixed!

dragonfly/implementations/at_operator.py
yadiw: Fixed.

dragonfly/implementations/at_operator.py
yadiw: I don't think so - I put this line here to hide(?) an unnecessary log produced by googleapiclient, which is supposed to have no impact on the Operator logger. And removing this line doesn't stop the duplication.

And so far I have no idea why the duplication would happen :(((

dragonfly/implementations/at_operator.py
yadiw: I tried to access the dict and it gave me a really long one - and it is supposed to contain all the channels in the workspalce :((( Maybe just try again...?

dragonfly/implementations/at_operator.py
yadiw: Fixed :)

dragonfly/implementations/at_operator.py
yadiw: By the time it reaches the else branch the `monitor_channel_id` is supposed to be a blank string - it is not assigned any value until line 393.

yadiw:
> Flaw in execution:
> 
> the dripline "service" can crash but the program doesn't crash, this breaks all dripline communication while the slack + GCal bits keep working; is this because of the while/try loop construction? running as daemon we don't need the keyboard interrupt anyway

I just removed the try & catch keyboard interrupt part and changed it to a try & finally - I think it would be better to send a slack message telling people it is stopped rather than being terminated silently...? 
I'm not sure about the first part though - do you mind to elaborate more on that?

> I'm confused because this looks like magic:
> 
> when I test the value of self.monitor_channel_id, it's a blank string; I don't understand how it correctly routes its operator shift change messages given this is the case

it does have a non-blank value later when I test it... So maybe try to get its value right before entering the loop?

dragonfly/implementations/at_operator.py
wcpettus: maybe something went sideways in startup when I was testing it
cannot reproduce is good with me, let's mark this resolved

dragonfly/implementations/at_operator.py
```diff
@@ -0,0 +1,470 @@
...
+        '''
+        Send a Slack message to @ the operator (@ temporary operator(s) if exists, otherwise @ the one listed on google calendar.)
+        channel: id of the channel where the message will be sent
+        '''
+        if len(self.temporary_operator_id) != 0:
+            logger.info('Found temporary operator(s)!')
+            message = ""
+            for operator_id in self.temporary_operator_id:
+                message += "<@" + operator_id + "> "
+            self.send_message(channel, message)
```
wcpettus: This implementation means if a temp operator is available, the normal operator is bypassed

I propose changing this to match the current "old" behavior - the "current_operator" is always paged, if temp operator(s) exist they are also paged

buzinsky:
The script (as of the version on Friday) successfully found the change in operators over the weekend, then on Monday.

dragonfly/implementations/at_operator.py
yadiw: Fixed :)

buzinsky:
The code is working for normal users, on slack, though there are still issues on the dragonfly side. 

1) `dragonfly get at_operator_interface.monitor_channel_id -b gus.p8` produces a blank string. I went into code, added a print(self.monitor_channel_id) if you set tempoperator, which it does successfully, so it is set. I'm not sure dragonfly is getting the values of the public variables after class initialization. Not sure why.
2) `dragonfly get at_operator_interface.update_interval -b gus.p8` crashes, dragonfly pinging of service because public variable is not json serializeable.  This still happens. When I call this, the bot responds normally on slack, though dragonfly commands result in:

at_operator_interface.current_operator_name(ret:102): []-> {'values': [None]}
2019-03-13T11:15:02[Level 25] dragonfly.subcommands.dripline_agent(87) -> return message: message not deliverable
for all variables.

wcpettus:
I'm not too worried about the `update_interval` being unavailable - this happens throughout our code whenever we have a non-serializable variable, and there's no convention for dealing with it.

If we think it's important to get this variable, we can just wrap it where `_update_interval` is the datetime.timedelta object and `update_interval` is a string representation, or # of seconds, or something.

dragonfly/implementations/at_operator.py
```diff
@@ -0,0 +1,472 @@
...
+    from google.oauth2.credentials import Credentials
+except ImportError:
+    pass
+
+import datetime, logging, os, time, json
+# for slack
+import slackclient
+
+from dripline.core import Endpoint, fancy_doc
+from .subprocess_mixin import *
```
wcpettus: do we use anything other than SlowSubprocessMixin from here? is an explicit import better?

dragonfly/implementations/at_operator.py
```diff
@@ -0,0 +1,472 @@
...
+        '''
+        imports = ['dateutil', 'funcsigs', 'googleapiclient', 'Credentials']
+        for import_module in imports:
+            if not import_module in globals():
+                raise ImportError(import_module + ' not found, required for AtOperator class.')
+
+        this_home = os.path.expanduser('~')
+        self.authentication_path = os.path.join(this_home, authentication_path)
+        try:
+            json.loads(open(self.authentication_path).read())
```
wcpettus: curiosity: is there any functional difference between this and `json.load(open(self.authentication_path))`?

dragonfly/implementations/at_operator.py
```diff
@@ -0,0 +1,472 @@
...
+        '''
+        Set up the slack client.
+        '''
+        slack = ''
+        config_file = json.loads(open(self.authentication_path).read())
+        if 'slack' in config_file and 'operator' in config_file['slack']:
+            slack = config_file['slack']['operator']
+            self.slack_client = slackclient.SlackClient(slack)
+        else:
+            logger.critical('Unable to find slack credentials for operator bot in <~/.project8_authentications.json>')
```
wcpettus: since the authentication_path is configurable, this should probably use .format(self.authentication_path)

dragonfly/implementations/at_operator.py
```diff
@@ -0,0 +1,472 @@
...
+
+    def get_calendar_credentials(self):
+        '''
+        Return google calendar credentials.
+        '''
+        creds_data = {}
+        config_file = json.loads(open(self.authentication_path).read())
+        if 'google' in config_file and 'calendar_credentials' in config_file['google']:
+            creds_data = config_file['google']['calendar_credentials']
+        try:
```
wcpettus: use parallel construction to get_slack_client above?  is if/else better, or if/try/except?

dragonfly/implementations/at_operator.py
yadiw: I think both of them would work in this case...?

dragonfly/implementations/at_operator.py
yadiw: fixed.

dragonfly/implementations/at_operator.py
yadiw: I feel if/try/except is better - the fact that `config_file['google']['calendar_credentials']` exists does not guarantee that it is a valid one. And the try/except can catch that.

PR #179, opened 2019-03-21
add sleep option to multido to allow time before check

wcpettus:
Purpose of this PR is to add a 'delay_check' option to every entry in a MultiDo, which adds a sleep between the set and the check to fix the power supply output bounce.  I also tried to clarify the error messages to make it easier to tell what fails without diving into the logs by saying which endpoint of a multido failed.

### DIFF
--- a/dragonfly/implementations/multido.py
+++ b/dragonfly/implementations/multido.py
@@ -172,7 +172,7 @@ class MultiDo(Endpoint):
 
             # if the value we are checking is a float/int
             if isinstance(value_get, (int,float)):
-                if not isinstance(target_value, (int,float)):
+                if not isinstance(target_value, (int, float)):
                     try:
                         target_value = float(target_value)
                     except ValueError:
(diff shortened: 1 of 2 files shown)


dragonfly/implementations/multido.py
```diff
@@ -162,7 +168,7 @@ def on_set(self, value):
                 logger.debug('no target_value given: using value ({}) as a target_value to check'.format(value))
                 target_value = value
             if value_get==None:
```
laroque: not a line you changed, but you should not check `==None`, should be `if value_get is None`

wcpettus: good catch, that shows up a couple times, now fixed

laroque:
did you do an integration test with the device and see it work? if so I'm happy

wcpettus:
I tested it live against the bouncing power supplies and verified that it (a) worked and (b) fixed the known issue.

PR #186, opened 2019-05-06
cleanup hornet, subprocess mixin

wcpettus:
- hornet only spams notices if files move
- subprocess mixin allows the control to be restarted


wcpettus:
I tested starting and stopping live to get the new behavior right, but did
not move files because that shouldn't change.

PR #195, opened 2019-06-06
removed channel id check in _check_psyllid_instance

cclaessens:


wcpettus:
Doesn't it need a check back in that the status return is == 4?  Since that's what active_channels dict return has as a criteria.  Unless the psyllid status is checked again elsewhere.

wcpettus:
Christine points out that this is checked next in _do_checks; because _check_psyllid_instance is a generic method (also used in prepare_daq_system), this is the division we want.

Ready to merge, will do this later.

Issue #194, opened 2019-06-06
remove roach DAQ threshold cache


wcpettus:
The stored_threshold of the roach_daq_run_interface doesn't seem to serve any purpose.  If the user wants to know the threshold, psyllid can be queried directly.  Psyllid now starts up with the correct trigger settings, so this only catches a mismatch when (a) the roach daq service restarts or (b) the psyllid threshold is directly set without dragonfly.

If (a) is the goal, why not create a variable which is more clearly named, `self.has_reset` which init's to True and is set to False when the daq system is prepared or something.  Or make sure that the ROACH DAQ service isn't stateful and doesn't care if it resets (which is probably already true?)

And (b) seems like an acceptable user choice, to directly set psyllid without using the convenience method.

cclaessens:
Having a `self.has_reset` seems like another state variable that doesn't automatically change if psyllid re-activates. The re-activation is checked by comparing the central frequencies of the roach and psyllid.

The threshold mismatch at the moment is not necessary anymore, because psyllid has the same default trigger settings. And I agree that applying the trigger settings via the channel daq interface is redundant now. 
But I disagree that (b) is an acceptable option. Most people are probably not aware of how to do the trigger settings in psyllid correctly, and the channel daq interface takes care of that for the operators.
In addition, if we ever want to take data with a different trigger configuration (and I am sure we will for some reason), then how do we prevent that psyllid restarts and continues the next run with an undesired trigger configuration, if we remove this check?

I would even say that maybe we need to do more trigger setting checks. It used to be enough to check one threshold, because psyllids default threshold was so unrealistic that something had to be wrong if a run was started with this threshold still set. Now, a run specific trigger configuration could have the same thresholds but other time window settings for example.

I am not sure how we should solve this.

wcpettus:
There are four possible restarts that could give rise to different failure modes, and will need different catches:

- psyllid
  - psyllid restarts with the wrong central frequency, defaulting to 50 MHz
    - the ROACH will never use a CF that low, so this will always flag the psyllid restart
    - psyllid needs the CF because it loads into the egg header, and this is an effective flag of restart, even though it is stateful
  - psyllid restarts with the correct default trigger setting (since we changed config file)
    - if we're running a special trigger, the CF check will alert to a psyllid restart, but we'll need to be careful to properly sync them in our script
- roach_daq_run_interface crash
  - what important state information is actually cached here?
  - currently the threshold variable is a catch, but I don't think we care?
- roach2_interface
[...]

PR #196, opened 2019-06-11
removed saved threshold and threshold check

cclaessens:


wcpettus:
Linking this to issue #194

PR #200, opened 2019-07-23
Feature/on call bots

buzinsky:
Changes to allow for general instances of operator bot (namely t2operator and expert)
Test seems to be working so far (2 days as of writing, handled shift transition)

If anything, the help messages could be improved, feedback appreciated.

### DIFF
--- a/dragonfly/implementations/__init__.py
+++ b/dragonfly/implementations/__init__.py
@@ -5,7 +5,7 @@ Implementation of instrument services.
 from __future__ import absolute_import
 
 from .alert_spammer import *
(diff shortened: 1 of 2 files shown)


wcpettus:
Looks good, if I ever touch this code I will probably mess up where the variable are all `on_call` and the method names use `oncall`.  Will probably get better advice on the helper function with this deployed and used.

buzinsky:
I was torn on that. I do prefer the underscores, but historically people call functions like `whoisop` or `tempoperator`. I did not have the courage to switch.

PR #204, opened 2019-09-10
allow specifying the binding key

laroque:
Another place where we hard coded the binding keys in a way that causes problems. Also better error handling on poorly configured named sensors.

### DIFF
--- a/dragonfly/implementations/sensor_logger.py
+++ b/dragonfly/implementations/sensor_logger.py
@@ -21,17 +21,18 @@ class SensorLogger(Gogol, PostgreSQLInterface):
     '''
     Project 8 slow control table implementation
     '''
-    def __init__(self, sensor_type_map_table, sensor_type_column_name='type', sensor_type_match_column='endpoint_name', data_tables_dict={}, **kwargs):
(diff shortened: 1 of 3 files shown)


dragonfly/implementations/sensor_logger.py
```diff
@@ -49,10 +50,14 @@ def add_endpoint(self, endpoint):
 
 
     def this_consume(self, message, basic_deliver):
+        logger.debug("consuming message to: {}".format(basic_deliver.routing_key))
         ### Get the sensor name
+        if not basic_deliver.routing_key.split('.')[0] == self.prefix:
+            logger.warning("should not consume this message")
+            return
```
wcpettus: I don't understand what this catches - if it's consuming isn't it on a bound queue and thus starting with the prefix?  This is more out of curiosity, because I know you've tested this.

laroque: I think it is still binding `#` from somewhere that I didn't find (or maybe I changed this then fixed that and forgot to revert this)... going to take another look at this and something else that came up over the weekend.

laroque: that was true when I added the lines, but should not be true now.

PR #207, opened 2019-10-09
Feature/nfs dung

buzinsky:
Removes .nfs files older than max_age
Tested on local directory, has expected behavior

I piggy-back over the max_age to delete directories, not creating a new user parameter (this may cause excess warnings, when dungbeetle will get around to deleting the bad files on the next cycle, but I haven't been able to reproduce it)

### DIFF
--- a/dragonfly/implementations/dungbeetle.py
+++ b/dragonfly/implementations/dungbeetle.py
@@ -14,7 +14,7 @@ class DungBeetle(Endpoint,Scheduler):
     '''
     def __init__(self,
                  root_dirs = [],
-                 max_age = {"minutes":3},
+                 max_age = {"hours":2},
                  ignore_dirs = [],
                  warning_interval = 12,
                  **kwargs):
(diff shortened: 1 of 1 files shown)


dragonfly/implementations/dungbeetle.py
```diff
@@ -33,9 +33,22 @@ def __init__(self,
...
 
+    # delete old .nfs files that get stranded 
+    def clean_nfs(self, path, min_creation_time):
+        creation_time = datetime.datetime.fromtimestamp(os.path.getctime(path))
+        if creation_time < min_creation_time and (not path in self.ignore_dirs):
+            try:
+                os.remove(path)
+                logger.info("nfs file [{}] has been removed.".format(path))
+            except:
+                pass
```
laroque: We should be more precise in the exception we're catching here and what we do in such a case. Probably `OSError` like below and then a critical alert that prints the message from that error so a human is notified to investigate.

PR #209, opened 2019-11-15
Feature/more generic logger

laroque:
Migrating some changes from ADMX:
## expected behavior fix(es)
- the sensor logger hard codes a requirement of using only specified payload keys, it should be possible to change this (seems better suited in the "table" class). Maybe even better would be an smaller helper method with only the logic for parsing the incoming payload data into an insert dictionary so that we can isolate and override just that logic.

## new feature
- table should support option for an "upsert" (insert or update if missing), we don't (currently) need it for P8, but it is a reasonable feature we may want access to and it is already implemented.

### DIFF
--- a/dragonfly/implementations/postgres_interface.py
+++ b/dragonfly/implementations/postgres_interface.py
@@ -31,10 +31,8 @@ import logging
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 
-__all__.append('PostgreSQLInterface')
-__all__.append("SQLTable")
-__all__.append('SQLSnapshot')
 
+__all__.append('PostgreSQLInterface')
 @fancy_doc
 class PostgreSQLInterface(Provider):
     '''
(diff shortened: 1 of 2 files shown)


laroque:
In the admx version I have this block in `_insert_with_return`:
```
         except Exception as err:                                                                                                                    
             if str(err).startswith('(psycopg2.IntegrityError)'):                                                                                    
                 raise DriplineDatabaseError(str(err))                                                                                               
             else:                                                                                                                                   
                 logger.critical('received an unexpected SQL error while trying to insert:\n{}'.format(str(ins) % insert_kv_dict))                   
                 logger.info('traceback is:\n{}'.format(traceback.format_exc()))                                                                     
                 raise
```
I don't understand why our version has a block `except Exception as err:` that doesn't `raise`, couldn't we do something like the above (assuming neither psycopg2 nor sqlalchemy provides a base exception we could handle), to only ignore other database exceptions? Is the point that if we have an error on one database insert, we want want it to break logging future messages? In that case I'd think we'd want the try/except way back in the spectrum_logger's `this_consume` method, with a single critical for any message that fails to log.

laroque:
We discussed the above question by phone. We agree that it is questionable at best to catch `Exception` and then print but not raise anything. This is catching way too much. We weren't sure what line to draw though, so we left the code as is and hope to re-consider either when we get around to a k8s migration, or at the time of a dripline v3 migration (hopefully then we'll be able to set a more concrete guideline on when things should continue to run but be loud, vs crash and let process monitoring figure it out.

laroque:
Note: this is revered, should have gone into develop

PR #218, opened 2025-02-06
Feature/chiller integration

renereimann:
Our chiller speaks some very special protocal, thus we implement speciallized Service and Entities. 
The following classes are implemented
* EthernetThermoFisherService
* ThermoFisherGetHexEntity
* ThermoFisherGetEntity

**EthernetThermoFisherService** implements the communication protocol with the device. 
A message consists of a header + command + number of data bytes + data bytes + checksum as indicated in this image from the manual 
image
The Service takes care of the lead_character, the lsb and msb address as well as byte counting and checksum calculation.
The communication is a half-duplex communication via a serial socket. The answer message follows the same structure as the request message where the command is repeated but the data bytes may be different. The resulting message is striped to data bytes and results are returned as hex-strings.

**ThermoFisherGetHexEntity** implements a get method. A cmd_str is interpreted as hex string. Some formating is needed since depending on the value the config file passes ints or strings. The resulting data from the device are send as hex-string. This entity can be used e.g. for status checking since the status is encoded in a bit-pattern. 

**ThermoFisherGetEntity** implements a get method that returns a numeric value. All get commands on the thermo fisher device return three bytes, represented as hex-string. The first byte encodes the unit and a decimal qualifier. The second and third bytes represent a 2byte int. The resulting final value then is value*decimal qualifier in encoded units. The ThermoFisherGetEntity only returns the numeric value without returning the unit, however a lookup table is used on the fly.

### DIFF
--- a/dripline/extensions/ethernet_thermo_fisher_service.py
+++ b/dripline/extensions/ethernet_thermo_fisher_service.py
@@ -1,15 +1,12 @@
-import re
-import socket
-import threading
 import time
 
-from dripline.core import Service, ThrowReply
+from dripline.core import ThrowReply
+from dripline.implementations import EthernetSCPIService
 
 import logging
 logger = logging.getLogger(__name__)
 
 __all__ = []
-
 
 __all__.append('EthernetThermoFisherService')
 
@@ -28,7 +25,7 @@ def hexstr_to_int(value):
 def bytes_to_ints(value):
     return [byte for byte in value]
 
-class EthernetThermoFisherService(Service):
+class EthernetThermoFisherService(EthernetSCPIService):
     '''
     A fairly specific subclass of Service for connecting to ethernet-capable thermo fisher devices.
     In particular, devices must support a half-duplex serial communication with header information, variable length data-payload and a checksum.
@@ -47,53 +44,18 @@ class EthernetThermoFisherService(Service):
             socket_info (tuple or string): either socket.socket.connect argument tuple, or string that
                 parses into one.
         '''
-        Service.__init__(self, **kwargs)
-
-        if isinstance(socket_info, str):
-            logger.debug(f"Formatting socket_info: {socket_info}")
-            re_str = "\([\"'](\S+)[\"'], ?(\d+)\)"
-            (ip,port) = re.findall(re_str,socket_info)[0]
-            socket_info = (ip,int(port))
-
-        self.alock = threading.Lock()
-        self.socket = socket.socket()
-        self.socket_timeout = float(socket_timeout)
-        self.socket_info = socket_info
         self.lead_char = lead_char
         self.msb = msb
         self.lsb = lsb
-
-    def send_to_device(self, commands, **kwargs):
-        '''
-        Standard device access method to communicate with instrument.
-        NEVER RENAME THIS METHOD!
-
-        commands (list||None): list of command(s) to send to the instrument following (re)connection to the instrument, still must return a reply!
-                             : if impossible, set as None to skip
-        '''
-        if isinstance(commands, str):
-            commands = [commands]
-        self.alock.acquire()
-
-        try:
-            data = self._send_commands(commands)
-        except Exception as err:
-            logger.critical(str(err))
-            try:
-                data = self._send_commands(commands)
-                logger.critical("Query successful after ethernet connection recovered")
-            except socket.error as err: # simply trying to make it possible to catch the error below
-                logger.warning(f"socket.error <{err}> received, attempting reconnect")
-                logger.critical("Ethernet reconnect failed, dead socket")
-                raise ThrowReply('resource_error_connection', "Broken ethernet socket")
-            except Exception as err: ##TODO handle all exceptions, that seems questionable
-                logger.critical("Query failed after successful ethernet socket reconnect")
-                raise ThrowReply('resource_error_no_response', str(err))
-        finally:
-            self.alock.release()
-        to_return = ';'.join(data)
-        logger.debug(f"should return:\n{to_return}")
-        return to_return
+        EthernetSCPIService.__init__(self, 
+                                     socket_timeout=socket_timeout,
+                                     socket_info=socket_info,
+                                     cmd_at_reconnect=["00"],
+                                     reconnect_test="0001",
+                                     command_terminator='',
+                                     response_terminator='ignored',
+                                     reply_echo_cmd=False, 
+                                     **kwargs)
 
     def calculate_checksum(self, command_bytes):
         """
@@ -135,34 +97,19 @@ class EthernetThermoFisherService(Service):
 
         for command in commands:
             cmd = self._assemble_cmd(command)
-            issue = True
-            try:
-                self.socket.close()
-                self.socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM, )
-                self.socket.settimeout(self.socket_timeout)
-                self.socket.connect(self.socket_info)
-                logger.debug(f"Socket: {self.socket}")
 
-                logger.debug(f"sending: {cmd}")
-                self.socket.send(cmd)
-                logger.debug("Wait for responds")
-                time.sleep(0.6)
-                logger.debug("Read header")
-                header = self.socket.recv(5)
-                nBytes = int(header[-1])
-                logger.debug("Read data")
-                data = self.socket.recv(nBytes)
-                logger.debug("Read checksum")
-                check = self.socket.recv(1)
-                response = header + data + check
-                issue = False
-            except Exception as err:
-                logger.warning(f"While socket communication we recived an error: {err}")
-            finally:
-                self.socket.close()
-                logger.debug("Closed connection")
-            if issue:
-                raise ThrowReply("connection_error", "No connection to device")
+            logger.debug(f"sending: {cmd}")
+            self.socket.send(cmd)
+            logger.debug("Wait for responds")
+            time.sleep(0.6)
+            logger.debug("Read header")
+            header = self.socket.recv(5)
+            nBytes = int(header[-1])
+            logger.debug("Read data")
+            data = self.socket.recv(nBytes)
+            logger.debug("Read checksum")
+            check = self.socket.recv(1)
+            response = header + data + check
             logger.info(f"Recived: {response}")
             if not self.check_checksum(response):
                 raise ThrowReply("checksum_error", "Message has invalid checksum")
--- a/dripline/extensions/ethernet_thermo_fisher_service.py
+++ b/dripline/extensions/ethernet_thermo_fisher_service.py
@@ -1,0 +1,149 @@
+import re
+import socket
+import threading
+import time
+
+from dripline.core import Service, ThrowReply
+
+import logging
+logger = logging.getLogger(__name__)
+
+__all__ = []
+
+
+__all__.append('EthernetThermoFisherService')
+class EthernetThermoFisherService(Service):
+    '''
+    A fairly specific subclass of Service for connecting to ethernet-capable thermo fisher devices.
+    In particular, devices must support a half-duplex serial communication with header information, variable length data-payload and a checksum.
+    '''
+    def __init__(self,
+                 socket_timeout=10,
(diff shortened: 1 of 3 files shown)


dripline/extensions/thermo_fisher_endpoint.py
```diff
@@ -0,0 +1,82 @@
...
+    def on_get(self):
+        # setup cmd here
+        to_send = [self.cmd_str]
+        logger.debug(f'Send cmd in hexstr: {to_send[0]}')
+        result = self.service.send_to_device(to_send)
+        logger.debug(f'raw result is: {result}')
+
+        return result
+
+    def on_set(self, value):
```
wcpettus, line 39: I don't understand this - if set isn't supported, you shouldn't need anything after the ThrowReply

dripline/extensions/thermo_fisher_endpoint.py
```diff
@@ -0,0 +1,82 @@
...
+logger = logging.getLogger(__name__)
+
+__all__ = []
+__all__.append("ThermoFisherGetHexEntity")
+__all__.append("ThermoFisherGetEntity")
+
+
+class ThermoFisherGetHexEntity(Entity):
+    '''
+    A simple endpoint which stores a value but returns that value + a random offset
```
wcpettus, line 13: Probably not an accurate description anymore

dripline/extensions/thermo_fisher_endpoint.py
```diff
@@ -0,0 +1,82 @@
...
+    def on_get(self):
+        # setup cmd here
+        to_send = [self.cmd_str]
+        logger.debug(f'Send cmd in hexstr: {to_send[0]}')
+        result = self.service.send_to_device(to_send)
+        logger.debug(f'raw result is: {result}')
+ 
+        # do something here
+        decimal = 10.**(-int(result[0], 16))
+        unit = self.units[int(result[1], 16)]
```
wcpettus, line 77: for now unit isn't used?

dripline/extensions/thermo_fisher_endpoint.py
```diff
@@ -0,0 +1,82 @@
...
+    def on_set(self, value):       
+        raise ThrowReply('message_error_invalid_method', f"endpoint '{self.name}' does not support set")
+        to_send = [cmd]
+        result = self.service.send_to_device(to_send)
+        logger.debug(f'raw result is: {result}')
+        # do something here
+        return result
+
+
+class ThermoFisherGetEntity(ThermoFisherGetHexEntity):
```
wcpettus, line 48: without seeing how these will be used, this name feels like it could be more descriptive - seeing a "GetEntity" and a "GetHexEntity" I might guess that this was more generic, rather than applying an additional correction to treat floats properly

dripline/extensions/thermo_fisher_endpoint.py
```diff
@@ -0,0 +1,82 @@
...
+
+import logging
+logger = logging.getLogger(__name__)
+
+__all__ = []
+__all__.append("ThermoFisherGetHexEntity")
+__all__.append("ThermoFisherGetEntity")
+
+
+class ThermoFisherGetHexEntity(Entity):
```
wcpettus, line 11: possibly more clear to change order to `ThermoFisherHexGetEntity` (since in other implementations there's never been another modifier between "Get" and "Entity")

dripline/extensions/ethernet_thermo_fisher_service.py
```diff
@@ -0,0 +1,172 @@
...
+
+        self.alock = threading.Lock()
+        self.socket = socket.socket()
+        self.socket_timeout = float(socket_timeout)
+        self.socket_info = socket_info
+        self.lead_char = lead_char
+        self.msb = msb
+        self.lsb = lsb
+
+    def send_to_device(self, commands, **kwargs):
```
wcpettus, line 66: this is curiously close to the standard SCPI send_to_device, but doesn't exactly match
can you elaborate what is getting fixed?  and if we should consider porting some of these changes into the base class?

dripline/extensions/ethernet_thermo_fisher_service.py
```diff
@@ -0,0 +1,172 @@
...
+        commands (list||None): list of command(s) to send to the instrument following (re)connection to the instrument, still must return a reply!
+                             : if impossible, set as None to skip
+        '''
+        all_data=[]
+
+        for command in commands:
+            cmd = self._assemble_cmd(command)
+            issue = True
+            try:
+                self.socket.close()
```
wcpettus, line 140: is it necessary to flush the socket connection every time you want to communicate?

wcpettus (changes requested):
I left a few comments throughout based on my quick impressions.  Happy to discuss more.

dripline/extensions/ethernet_thermo_fisher_service.py
```diff
@@ -0,0 +1,172 @@
...
+def bytes_to_hexstr(value):
+    return value.hex()
+
+def hexstr_to_int(value):
+    return int(value, 16)
+
+def bytes_to_ints(value):
+    return [byte for byte in value]
+
+class EthernetThermoFisherService(Service):
```
nsoblath, line 31: I'm wondering if it would make sense to inherit from `dripline.EthernetSCPIService` and reimplement `_send_commands()` (we could think about restructuring `EthernetSCPIService._send_commands()` to use `_assemble_cmd()`, which would further simplify this class), and `_listen()`, plus adding the other necessary methods.

dripline/extensions/thermo_fisher_endpoint.py
```diff
@@ -0,0 +1,82 @@
...
+    '''
+    A simple endpoint which stores a value but returns that value + a random offset
+    '''
+
+    def __init__(self, 
+                 get_str=None,
+                 **kwargs):
+        '''
+        Args:
+            cmd_str: hexstring of the command, e.g. 20
```
nsoblath, line 21: Arg name in the comment should match the actual argument name

nsoblath (changes requested):

nsoblath (commented):
Thank you for the very thorough description of your feature and what it accomplishes with respect to communicating in this particular way with the Thermo Fisher devices.  That was a great description of the PR.

dripline/extensions/ethernet_thermo_fisher_service.py
```diff
@@ -0,0 +1,172 @@
...
+                             : if impossible, set as None to skip
+        '''
+        if isinstance(commands, str):
+            commands = [commands]
+        self.alock.acquire()
+
+        try:
+            data = self._send_commands(commands)
+        except Exception as err:
+            logger.critical(str(err))
```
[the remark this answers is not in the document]
wcpettus: There is no `_reconnect` in this implementation - all the socket connection is handled down in `_send_commands` (which is a separate question I have).  Maybe I just answered my own question about why you need to override the inherited method.
The original argument was that a socket.error means the ethernet device needs to flush the socket connection (=reconnect).  If it's a different kind of error, then I'm not sure what's going on and any troubleshooting is shooting in the dark.  Mostly the difference is the error messages you get out - which makes it a fairly quick structure in the original.

dripline/extensions/thermo_fisher_endpoint.py
renereimann: Unit is constant and does not change. We either can return f"{value} {unit}" as a string or we just return the value as numeric value. If I am right we can not have a tuple as payload. If its a string that contains the units as well, then you would need a calibration function that strips the unit again and makes it a float. I see pros and cons for both versions. As you may see I evaluate both, but just return the value. I had a version before where I had value and unit as a string.

dripline/extensions/thermo_fisher_endpoint.py
renereimann: What about GetNumericEntity ? That would also solve the question about value + unit string or just value?

dripline/extensions/thermo_fisher_endpoint.py
renereimann: Its an artefact from me starting with a copy of the standard SCPI servie. I will remove this.

dripline/extensions/thermo_fisher_endpoint.py
renereimann: switched the Hex and Get

dripline/extensions/ethernet_thermo_fisher_service.py
renereimann: I do not keep the connection open and the handling of the reconnection. Thats moved into the `_send_commands` function.

dripline/extensions/ethernet_thermo_fisher_service.py
renereimann: May be ;) . The chiller seems to drop connection every 10 minutes. So there are "frequent" reconnects. Also a open connection blocks the serial port and does not allow to speak to the device in another way. Maybe thats wanted but at least while testing I had several times the issue that something still was connected and I could not connect to the device. 

Thats more a general question I have and I looked through several forums: What is the advantage of keeping a serial connection open vs reconnecting at each communiction. In general it sounds the prefered way is to keep the serial connection open all time but I do not see the reason for that. Is it that its blocking other connections to the device?

dripline/extensions/ethernet_thermo_fisher_service.py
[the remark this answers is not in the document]
renereimann: The `_reconnect` function as implemented for the SCPI device also tests for the `cmd_at_reconnect`, which is not implemented here. In principle thats also possible for the thermo fisher device (there is a "request acknowledge") but the structure is the same as a ThermoFisherHexGetEntity implements and the result is a version number defined in hex string. So there has to be additional wrapping in here.

dripline/extensions/thermo_fisher_endpoint.py
renereimann: Fixed

dripline/extensions/ethernet_thermo_fisher_service.py
renereimann: If I revert back to use the `_reconnect` function I can inherit from EhternetSCPIService. I anyway have to overwrite the __init__ due to different arguments that are needed but then only _send_commands() would need modifications.

dripline/extensions/thermo_fisher_endpoint.py
```diff
@@ -0,0 +1,82 @@
...
+            cmd_str: hexstring of the command, e.g. 20 
+        '''
+        if get_str is None:
+            raise ValueError('<base_str is required to __init__ ThermoFisherGetEntity instance')
+        else:
+            self.cmd_str = str(get_str).zfill(2)
+        Entity.__init__(self, **kwargs)
+
+    @calibrate()
+    def on_get(self):
```
renereimann, line 30: I have a question on the `@calibrate` qualifier on the `on_get` method. At first I tried to call the `ThermoFischerHexGetEntity.on_get()` function within the `ThermoFischerNumericGetEntity.on_get()`, since we first have to get the hex string as in the ThermoFischerHexGetEntity and then do further work on it. However when running this code, I got errors from the calibration method, which tried to do the calibration on the result of `ThermoFischerHexGetEntity.on_get()` instead of the `ThermoFischerNumericGetEntity.on_get()`. Because of that I fully rewrote this piece. Any suggestions if that can be simplified?

dripline/extensions/thermo_fisher_endpoint.py
renereimann: updated

renereimann:
I now rewrote the EthernetThermoFischerService which now inherits from EthernetSCPIService. 
That made the class much simpler and also reverts back to make use of the _reconnect function.
Please review the changed implementation again. 

I have tested the implementation with our device and things are working.

nsoblath (changes requested):
Nice reorganization.  The main thing that I think needs to be changed is the way the arguments are handled in `EthernetThermoFisherService`.

nsoblath (approved):
Thanks for these updates, @renereimann

PR #219, opened 2025-02-06
initial muxer implementation

wcpettus:
Port of MuxerProvider and associated Entity class to dl3 version.

### DIFF
--- a/dripline/extensions/muxer_service.py
+++ b/dripline/extensions/muxer_service.py
@@ -1,0 +1,87 @@
+'''
+A class to interface with the multiplexer aka muxer instrument
+'''
+
+from dripline.core import ThrowReply, Entity, calibrate
+from dripline.implementations import EthernetSCPIService
+
+import logging
+logger = logging.getLogger(__name__)
+
+__all__ = []
+__all__.append('MuxerService')
+
+class MuxerService(EthernetSCPIService):
+    '''
+    Provider to interface with muxer
+    '''
+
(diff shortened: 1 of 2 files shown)


wcpettus:
@miniwynne muxer.yaml should be in the cca-controls repo
not sure if we have a good spot to park single-service test docker files

nsoblath (changes requested):
Overall looks quite good.  Can we please move the example .yaml files to an examples directory, or something like that?  This would be a good opportunity to add it so that the top directory doesn't get overwhelmed.

dripline/extensions/muxer_service.py
```diff
@@ -0,0 +1,149 @@
...
+    def on_get(self):
+        result = self.service.send_to_device([self.get_str.format(self.ch_number)])
+        logger.debug('very raw is: {}'.format(result))
+        return result.split()[0]
+
+    def on_set(self, value):
+        raise ThrowReply('message_error_invalid_method',
+                        f'endpoint {self.name} does not support set')
+__all__.append('MuxerRelay')
+class MuxerRelay(Entity):
```
wcpettus, line 89: switch inheritance to be from `FormatEntity`, see:
https://github.com/project8/dragonfly/blob/dl2/master/dragonfly/implementations/spime_endpoints.py#L683
https://github.com/project8/dragonfly/blob/dl2/master/dragonfly/implementations/spime_endpoints.py#L743

PR #221, opened 2025-05-28
CommandEntity and FormatEntity with asteval supports

renereimann:
We created two new endpoint entities
* CmdEntity can execute a command, e.g. degaussing of a pressure gauge. This is differentent from get and set
* FormatEntityAsteval which allows for a function to be defined in the config file that is used to convert the response (similar to calibration but more focussed on formating the response"

The Entities are used within the MATS setup for some time and seem to work fine.

### DIFF
--- a/dripline/extensions/asteval_endpoint.py
+++ b/dripline/extensions/asteval_endpoint.py
@@ -1,0 +1,75 @@
+import asteval # used for FormatEntity
+import re # used for FormatEntity
+
+from dripline.core import calibrate, ThrowReply
+from dripline.implementations import FormatEntity
+from dripline.core import Entity
+
+import logging
+logger = logging.getLogger(__name__)
+
+__all__ = []
+
+
+
+__all__.append('FormatEntityAsteval')
+class FormatEntityAsteval(FormatEntity):
+    '''
+    Utility Entity allowing arbitrary set and query syntax and formatting for more complicated usage cases
+    No assumption about SCPI communication syntax.
+    '''
+
+    def __init__(self,
+                 asteval_get_string="def f(response): return response",
+                 **kwargs):
+        '''
+        Args:
+            get_str (str): sent verbatim in the event of on_get; if None, getting of endpoint is disabled
+            get_reply_float (bool): apply special formatting to get return
+            set_str (str): sent as set_str.format(value) in the event of on_set; if None, setting of endpoint is disabled
+            set_value_lowercase (bool): default option to map all string set value to .lower()
+                **WARNING**: never set to False if using a set_value_map dict
+            set_value_map (str||dict): inverse of calibration to map raw set value to value sent; either a dictionary or an asteval-interpretable string
+            extract_raw_regex (str): regular expression search pattern applied to get return. Must be constructed with an extraction group keyed with the name "value_raw" (ie r'(?P<value_raw>)' ) 
+        '''
+        super().__init__(**kwargs)
+        self.asteval_get_string = asteval_get_string # has to contain a definition "def f(response): ... return value"
+        logger.debug(f'asteval_get_string: {repr(self.asteval_get_string)}')
+        self.evaluator(asteval_get_string)
+
+    @calibrate()
+    def on_get(self):
+        if self._get_str is None:
+            raise ThrowReply('message_error_invalid_method', f"endpoint '{self.name}' does not support get")
+        result = self.service.send_to_device([self._get_str])
+        logger.debug(f'result is: {result}')
+        if self._extract_raw_regex is not None:
+            first_result = result
+            matches = re.search(self._extract_raw_regex, first_result)
+            if matches is None:
+                logger.error('matching returned none')
+                # exceptions.DriplineValueError
+                raise ThrowReply('resource_error', 'device returned unparsable result, [{}] has no match to input regex [{}]'.format(first_result, self._extract_raw_regex))
+            logger.debug(f"matches are: {matches.groupdict()}")
+            result = matches.groupdict()['value_raw']
+        
+        result = result.replace('\x00', '')
+
+        processed_result = self.evaluator(f"f('{result}')")
+        logger.debug(f"processed_result: {repr(processed_result)}")
+        return processed_result
+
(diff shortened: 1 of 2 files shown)


dripline/extensions/asteval_endpoint.py
```diff
@@ -0,0 +1,75 @@
...
+            extract_raw_regex (str): regular expression search pattern applied to get return. Must be constructed with an extraction group keyed with the name "value_raw" (ie r'(?P<value_raw>)' ) 
+        '''
+        super().__init__(**kwargs)
+        self.asteval_get_string = asteval_get_string # has to contain a definition "def f(response): ... return value"
+        logger.debug(f'asteval_get_string: {repr(self.asteval_get_string)}')
+        self.evaluator(asteval_get_string)
+
+    @calibrate()
+    def on_get(self):
+        if self._get_str is None:
```
nsoblath, line 42: Should this block (lines 42-54) be replaced by a call to `FormatEntity.on_get()`?

dripline/extensions/asteval_endpoint.py
```diff
@@ -0,0 +1,75 @@
...
+        processed_result = self.evaluator(f"f('{result}')")
+        logger.debug(f"processed_result: {repr(processed_result)}")
+        return processed_result
+
+
+
+__all__.append('CmdEntity')
+class CmdEntity(Entity):
+    def __init__(self, cmd_str=None, **kwargs):
+        Entity.__init__(self, **kwargs)
```
nsoblath, line 67: Class and init function need documentation

dripline/extensions/asteval_endpoint.py
```diff
@@ -0,0 +1,75 @@
...
+        logger.debug(f"I get cmd_str: {cmd_str}, which is of type {type(cmd_str)}.")
+        self.cmd_str = cmd_str
+
+    def cmd(self):
+        logger.debug("Command function was successfully called")
+        if self.cmd_str is None:
+            raise ThrowReply('service_error', f"endpoint '{self.name}' does not support cmd")
+        return self.service.send_to_device([self.cmd_str])
```
nsoblath, line 75: The call to `Service.send_to_device()` makes me think this is intended to be used with an `EthernetSCPIService`.  That should be included in the documentation.

dripline/extensions/asteval_endpoint.py
```diff
@@ -0,0 +1,75 @@
...
+    Utility Entity allowing arbitrary set and query syntax and formatting for more complicated usage cases
+    No assumption about SCPI communication syntax.
+    '''
+
+    def __init__(self,
+                 asteval_get_string="def f(response): return response",
+                 **kwargs):
+        '''
+        Args:
+            get_str (str): sent verbatim in the event of on_get; if None, getting of endpoint is disabled
```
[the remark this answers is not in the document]
renereimann: I added the additional asteval_str in the documentation. I still keep the other arguments since they are propergated via **kwargs and are needed by the baseclass

dripline/extensions/asteval_endpoint.py
renereimann: Added that this is for EthernetSCPIService

wcpettus:
Two thoughts/questions on the CmdEntity related to this PR, but beyond the scope:
- dripline cmd supports additional arguments, but the CmdEntity as written doesn't use them - if we get a device where it would be useful, could consider how best to flexibly incorporate the optional arguments (although I'm not thinking of an obvious application - since often you would just want a set there instead)
- in the example config, you have an (empty) calibration - this isn't used here, right? (config file is not the PR, so no effect)

wcpettus (changes requested):
If I post a review, do my other comments go from "Pending" to be visible?

dripline/extensions/asteval_endpoint.py
```diff
@@ -0,0 +1,71 @@
...
+        Args:
+            cmd_str (str): sent verbatim in the event of cmd().
+        '''
+        Entity.__init__(self, **kwargs)
+        logger.debug(f"I get cmd_str: {cmd_str}, which is of type {type(cmd_str)}.")
+        self.cmd_str = cmd_str
+
+    def cmd(self):
+        logger.debug("Command function was successfully called")
+        if self.cmd_str is None:
```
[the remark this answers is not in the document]
renereimann: Good point, I moved it to the constructor.

dripline/extensions/asteval_endpoint.py
```diff
@@ -0,0 +1,75 @@
...
+            first_result = result
+            matches = re.search(self._extract_raw_regex, first_result)
+            if matches is None:
+                logger.error('matching returned none')
+                # exceptions.DriplineValueError
+                raise ThrowReply('resource_error', 'device returned unparsable result, [{}] has no match to input regex [{}]'.format(first_result, self._extract_raw_regex))
+            logger.debug(f"matches are: {matches.groupdict()}")
+            result = matches.groupdict()['value_raw']
+        
+        result = result.replace('\x00', '')
```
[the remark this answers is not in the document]
renereimann: I removed it. This is not needed

dripline/extensions/cmd_endpoint.py
```diff
@@ -0,0 +1,31 @@
+import re # used for FormatEntity
+
+from dripline.core import calibrate, ThrowReply
```
wcpettus: unused

dripline/extensions/cmd_endpoint.py
```diff
@@ -0,0 +1,31 @@
+import re # used for FormatEntity
```
wcpettus: unused

dripline/extensions/asteval_endpoint.py
```diff
@@ -0,0 +1,40 @@
+import asteval # used for FormatEntity
+import re # used for FormatEntity
```
wcpettus: asteval nor re used

wcpettus (approved):
Looks good - we should keep an eye on CmdEntity, that may be worth elevating into dl-py codebase

PR #222, opened 2025-07-08
Arduino Service and Endpoint

renereimann:
We have several Arduinos that run some sensors and we need to comminicate with.
Since we do not have full SCIPY implementation this is a light weighted version of an endpoint and service.
This code is nicely running at our setup.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 ARG img_user=ghcr.io/driplineorg
 ARG img_repo=dripline-python
-#ARG img_tag=develop-dev
-ARG img_tag=receiver-test
+ARG img_tag=v5.1.0
 
 FROM ${img_user}/${img_repo}:${img_tag}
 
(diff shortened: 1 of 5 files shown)


wcpettus:
If I'm understanding this correctly, the EthernetService part of this code is essentially a stripped-back version of the standard EthernetSCPIService that allows for socket instability, and thus flushes the socket connection every time it communicates.

Some assorted thoughts this raises (which are not prerequisites for this PR, but maybe to start a discussion that might need to migrate over to an issue thread):
- What are the relative merits of persistent vs continually-reset socket connections?
  - There's a lot of variance to how devices handle ethernet connections - some are single connection, some support multiple; some break the connection if you try to manually control; this can afford some system state guarantee because you know the SC system is the only one communicating with a device?
- Is it possible to make Arduino's persistent socket compliant?
  - Years ago I played around with a few standard Arduino's, and found a significant drop rate requiring a socket refresh.  It was true across all models that used WiFi, but when I used a device with built-in ethernet everything was fine (that's why the Phase II lab used a bunch of ESP32-Gateway devices by Olimex).
- Are we following best practice with our socket connections?
  - Prompted by this PR, I poked around a bit, and found some pages like this recommending the use of socket.shutdown in the workflow (https://www.pythonlore.com/closing-and-releasing-resources-with-socket-close/).  Would need to read more.

wcpettus (commented):
Adding review to post code comments.

PR #223, opened 2025-07-08
modbus service and entity

renereimann:
Implementation of a modbus service and entity using the pymodbus module.
This implementation is working with our backout controller.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -19,7 +19,7 @@ jobs:
   test_docker:
     name: Build and Test in Docker
 
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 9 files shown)


renereimann:
I finally found a way to make a proper diff between the two implementations. The issue was that the files lived in different directories and thus were not directly compared against each other.

Looking at the diff there are a series of difference. Most likely the issue why the old version did not work for us was because it was operating on input registers while we need operation on holding registers. However the old code is easily extendable and has some decoding / encoding of the payload in addition. 

I will merge these two versions once the device is setup, so I can test the implementation.

renereimann:
All functionality in here is now implemented in #227 . Thus we can close this pull request.

PR #224, opened 2025-07-29
Pfeiffer Endpoint: Implementation of Pfeiffer telegram protocol

renereimann:
Implementation of Pfeiffer telegram protocol following documentation from Pfeiffer manuals. 
The frame of the protocol is defined by:
pfeiffer_telegram
Each parameter has a number (which may differ from device to device). For values Pfeiffer has a table of predefined datatypes and how they have to be formated:
pfeiffer_datatypes

Implementation works very nice with our pressure gauge of type HPT 200 from Pfeiffer. 
The configuration file significantly improved in readability (see example folder).

### DIFF
--- a/dripline/extensions/pfeiffer_endpoint.py
+++ b/dripline/extensions/pfeiffer_endpoint.py
@@ -13,7 +13,17 @@ class PfeifferEntity(Entity):
     Conversion of datatypes and checksum checking are implemented.
     '''
 
-    _datatype = {"bool_old": 6, "uint": 6, "ureal": 6, "string": 6, "bool": 1, "ushort": 3, "uexpo": 6, "str16": 16, "str8": 8, "query": 2}
+    _datatype = {"bool_old": 6, 
+                 "uint":     6, 
+                 "ureal":    6, 
+                 "string":   6, 
+                 "bool":     1, 
+                 "ushort":   3, 
+                 "uexpo":    6, 
+                 "str16":   16, 
+                 "str8":     8, 
+                 "query":    2
+                 }
     
     def __init__(self,
                  parameter = 303,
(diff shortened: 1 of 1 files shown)


pkolbeck (approved):
Is it true that the response datatype is always the same as the command datatype?

dripline/extensions/pfeiffer_endpoint.py
```diff
@@ -0,0 +1,140 @@
...
+        '''
+        Entity.__init__(self, **kwargs)
+        self.parameter = parameter
+        self.datatype = datatype
+        self.unit_address = unit_address
+
+    def get_checksum(self, string):
+        return sum([ord(c) for c in string])%256
+
+    def format_value(self, value):
```
nsoblath: This function doesn't have a default return for the situation where none of the if statements are satisfied.

dripline/extensions/pfeiffer_endpoint.py
```diff
@@ -0,0 +1,140 @@
...
+            return int(value)/100.
+        if self.datatype == "uexpo":
+            return int(value[:4])/1000. * 10**(int(value[4:]) - 20)
+
+    def disensemble_result(self, reply):
+        if self.get_checksum(reply[:-3]) != int(reply[-3:]):
+            logger.debug("checksum not matching")
+        address = int(reply[:3])
+        action = reply[3:5]
+        parameter = int(reply[5:8])
```
wcpettus: comment lines (or just add this mapping to the method comment for future reference) for address, action, and parameter - unused variables?

wcpettus (changes requested):
Looks good, a couple small comments.

PR #227, opened 2025-09-22
general modbus entity/service 

Robbie-1208:


dripline/extensions/modbus/ethernet_modbus_service.py
```diff
@@ -0,0 +1,160 @@
+try:
+    import pymodbus
+    from pymodbus.client import ModbusTcpClient
+    from pymodbus.payload import BinaryPayloadDecoder
```
renereimann: Is no longer available in pymodbus version 3.11.4. I do not see explicitly to which version is used here.

dripline/extensions/modbus/ethernet_modbus_service.py
```diff
@@ -0,0 +1,160 @@
...
+__all__.append('EthernetModbusService')
+class EthernetModbusService(Service):
+    '''
+    Service for connectivity to ModbusTCP instruments built on pymodbus library.
+    '''
+    def __init__(self,
+                 ip_address,
+                 indexing='protocol',
+                 wordorder='big',
+                 byteorder='big',
```
renereimann: The byte order is fixed by the modbus standard see https://github.com/pymodbus-dev/pymodbus/pull/2456

dripline/extensions/modbus/ethernet_modbus_service.py
```diff
@@ -0,0 +1,160 @@
...
+
+__all__.append('EthernetModbusService')
+class EthernetModbusService(Service):
+    '''
+    Service for connectivity to ModbusTCP instruments built on pymodbus library.
+    '''
+    def __init__(self,
+                 ip_address,
+                 indexing='protocol',
+                 wordorder='big',
```
renereimann: The wordorder has to be specified per entity not per service.

dripline/extensions/modbus/ethernet_modbus_service.py
```diff
@@ -0,0 +1,160 @@
...
+            logger.info('Device returned {}'.format(result.registers))
+
+        except Exception as e: 
+            logger.debug(f'read registers failed: {e}. Attempting reconnect.')
+            self._reconnect()
+
+        if n_reg == 1:
+            return result[0]
+        else:
+            return BinaryPayloadDecoder.fromRegisters(result.registers, wordorder=self.word, byteorder=self.byte)
```
renereimann: The BinaryPayloadDecoder is removed from current versions of pymodbus. The conversion has to happen on the per Entity and not on the Service level.

dripline/extensions/modbus/ethernet_modbus_service.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def write_register(self, register, value):
+        '''
+        This register only works with reg_type = 0x10
+        '''
+        logger.debug('writing {} to register {}'.format(value, register))   
+
+        if not isinstance(value, list):
```
renereimann: If the value is not a list we should use reg_type= 0x06 which writes exactly one register. We should use self.client.write_register(register + self.offset, value)

dripline/extensions/modbus/ethernet_modbus_service.py
```diff
@@ -0,0 +1,160 @@
...
+        self.register = register
+        self.n_reg = n_reg
+        self.data_type = data_type
+        Entity.__init__(self, **kwargs)
+
+    @calibrate()
+    def on_get(self):
+        result = self.service.read_register(self.register, self.n_reg)
+        if self.data_type == 'float32':
+            result = result.decode_32bit_float()
```
renereimann: Use `convert_from_registers` instead.
This can be done by doing
`self.client.convert_from_registers(result.registers, pymodbus.client.DATATYPE.FLOAT32, word_order="big")`

wcpettus:
So you can only define the endian-ness of the word order and not the byte order now?  I'm hoping this doesn't cause issues for any devices down the road.  The documentation on the new method seems thin, with a lot of the docs still using BinaryPayloadDecoder despite the deprecation warnings, but that's a developer side issue.

Our device is offline right now, so can't test.  Will have to work out any issues on a hot fix.  I may go back and allow endianness to be defined at a Service level, since that should be a device property and not require specification at the individual entity level.

dripline/extensions/ethernet_modbus_service.py
```diff
@@ -0,0 +1,199 @@
+try:
+    import pymodbus
+    from pymodbus.client import ModbusTcpClient
+except ImportError:
+    pass
+
+import scarab
```
wcpettus: remove import

dripline/extensions/ethernet_modbus_service.py
```diff
@@ -0,0 +1,199 @@
...
+
+        try:
+            if isinstance(value, list):
+                return self.client.write_registers(register + self.offset, value).registers
+            else:
+                return self.client.write_register(register + self.offset, value).registers[0]
+        except Exception as e:
+            logger.debug(f'write_registers failed: {e}. Attempting reconnect.')
+            self._reconnect()
+            try:
```
wcpettus: same as with reads above

PR #233, opened 2026-02-06
Feature/p8 sensor logger

renereimann:
Implementation of a sensor logger matching the database layout of project8.
The endpoint_name is used to check in the id_map table which type of data the endpoint is. 
In a second step the correct table is selected and data are inserted.

### DIFF
--- a/dripline/extensions/__init__.py
+++ b/dripline/extensions/__init__.py
@@ -13,3 +13,4 @@ from .asteval_endpoint import *
 from .thermo_fisher_endpoint import *
 from .ethernet_thermo_fisher_service import *
 from .pfeiffer_endpoint import *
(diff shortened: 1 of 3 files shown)


wcpettus:
I attempted to tighten this by tweaking the base PostgresSensorLogger base class to add a bit of flexibility, then making a Mapped Logger class that could inherit expand on it (doing essentially the same thing).  That lives over in dl-py:
https://github.com/driplineorg/dripline-python/pull/223

My reasoning was:
- ADMX and various P8 setups want something like this, so there's demand for the expansion
- This method reduces code duplication for easier maintenance

Comments should follow of whether this was a good idea or not, but now the PRs are linked.

PR #235, opened 2026-03-02
Feature/Checklist Server

renereimann:
We implemented a flask server that provides checklist to check and monitor the conditions of the experiment on a regular basis.
Several checklists can be configured in a yaml config file. Supported are checkboxes and free text. Links to monitoring pages can be added optionally. Based on the filled out and submitted checklist an automatic entry in the elog is generated.

Overview over different checklists
Checklist_overview

Implementation of one checklist
Checklist_example

Resulting Elog Entry from Checklist
Checklist_elog


wcpettus:
This is a really neat utility
Does it belong in dragonfly?  As I understand it, this exists apart, doesn't touch the mesh, but is a standalone service that supports your lab operations.

renereimann:
@wcpettus Yess, its standalone and not touching the mesh. 
Maybe I did not got the scope right of what dragonfly is supposed to contain and what not.
If dragonfly just is meant to be for things interacting with the mesh, I may move this script to a different location.

pkolbeck:
Only repository I found that, conceptually, would support like this in the project8 repositories is "swarm": https://github.com/project8/swarm/tree/master
This seems to only be used for Go scripts and apps, of which several are no longer in use. We could use that repo to host something like this, change the scope of dragonfly, make a new repository that hosts useful in-lab scripts, flasks, etc., or we could make a directory in the P8 scripts repository? I think I would lean towards one of the latter two.

PR #236, opened 2026-03-11
Feature/alarm system v2

renereimann:
Adding additional features to the alarm system:

Modifications to watchdog.py

- Support an optional "enable" parameter for each check in check_endpoints. If the enable parameter is present and False, the check is skipped. This allows to quickly and temporarly disable the check e.g. if a device is powered off or is known to be in a bad state.
- The config file is automatically reloaded before each check_interval. This allows for modifications in the config file to take immidiete effects and the docker container does not have to be restarted if an additional check of an endpoint is added.
- The slack_hook is stored in a separate file. If the slack hook is exposed to git (even in private repos), slack / git detects the hook disables it due to security concerns. Thus we could not commit changes of the config file to git. By separating the slack_hook the main part can go into github.

An additional HTTP Server using the flask app is setup. The HTTP Server displays a table with all endpoints that are checked.
Web forms allow to Add, Delete and Update the configuration of checked endpoints (see picture below). In combination with the modifications to watchdog this allows to fast enabling / disabling and modification of the endpoint checks.

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ FROM ${img_user}/${img_repo}:${img_tag}
 COPY . /usr/local/src_dragonfly
 
 WORKDIR /usr/local/src_dragonfly
-RUN pip install docker pymodbus
+RUN pip install docker pymodbus Flask
 RUN pip install .
 
 WORKDIR /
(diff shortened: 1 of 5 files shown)


pkolbeck (approved):
This looks good to me. Design makes sense. Example configs would be at Mainz, including how the flask app is set up.

